PR #21, opened 2019-07-02
Compile with no implicit returns.

rictic:
I'm prototyping using this module as a plugin in https://github.com/bazelbuild/rules_typescript

As part of pulling it into google's internal repo for the prototyping, it has to compile with google's internal tsconfig, which includes noImplicitReturns. I'm not sure that it's an improvement in legibility, so if you don't want to merge this that's fine, we can use a patch set on top.

### DIFF
--- a/src/analyze/document-analyzer/html/definition/definition-for-html-attr.ts
+++ b/src/analyze/document-analyzer/html/definition/definition-for-html-attr.ts
@@ -20,4 +20,5 @@ export function definitionForHtmlAttr(htmlAttr: HtmlNodeAttr, { htmlStore, docum
 			target: target.declaration
 		};
 	}
+	return;
 }
--- a/src/analyze/document-analyzer/html/diagnostic/validate-html-node-attr-assignment.ts
+++ b/src/analyze/document-analyzer/html/diagnostic/validate-html-node-attr-assignment.ts
@@ -145,6 +145,7 @@ function validateHtmlAttrAssignmentRules(htmlAttr: HtmlNodeAttr, typeB: SimpleTy
 			}
 			break;
 	}
+	return;
 }
 
 function validateHtmlAttrSlotAssignment(htmlAttr: HtmlNodeAttr, { document, htmlStore, config, program }: LitAnalyzerRequest): LitHtmlDiagnostic[] | undefined {
@@ -239,6 +240,7 @@ function validateHtmlAttrAssignmentTypes(htmlAttr: HtmlNodeAttr, { typeA, typeB 
 			}
 		];
 	}
+	return;
 }
 
 function validateStringifiedAssignment(
@@ -389,6 +391,7 @@ function validateStringifiedAssignment(
 			];
 		}
 	}
+	return;
 }
 
 function validateHtmlAttrDirectiveAssignment(htmlAttr: HtmlNodeAttr, { typeA, typeB }: { typeA: SimpleType; typeB: SimpleType }, request: LitAnalyzerRequest): LitHtmlDiagnostic[] | undefined {
--- a/src/analyze/document-analyzer/html/diagnostic/validate-html-node-attr.ts
+++ b/src/analyze/document-analyzer/html/diagnostic/validate-html-node-attr.ts
@@ -121,4 +121,5 @@ function findSuggestedTarget(name: string, ...tests: Iterable<HtmlAttrTarget>[])
 			return match;
 		}
 	}
+	return;
 }
--- a/src/analyze/document-analyzer/html/lit-html-document-analyzer.ts
+++ b/src/analyze/document-analyzer/html/lit-html-document-analyzer.ts
@@ -76,6 +76,7 @@ export class LitHtmlDocumentAnalyzer {
 		} else if (isHTMLAttr(hit)) {
 			return definitionForHtmlAttr(hit, request);
 		}
+		return;
 	}
 
 	getRenameInfoAtOffset(document: HtmlDocument, offset: number, request: LitAnalyzerRequest): LitRenameInfo | undefined {
@@ -92,6 +93,7 @@ export class LitHtmlDocumentAnalyzer {
 				target: hit
 			};
 		}
+		return;
 	}
 
 	getRenameLocationsAtOffset(document: HtmlDocument, offset: number, request: LitAnalyzerRequest): LitRenameLocation[] {
@@ -109,6 +111,7 @@ export class LitHtmlDocumentAnalyzer {
 		if (isHTMLAttr(hit)) {
 			return quickInfoForHtmlAttr(hit, request);
 		}
+		return;
 	}
 
 	getOutliningSpans(document: HtmlDocument): LitOutliningSpan[] {
--- a/src/analyze/lit-analyzer.ts
+++ b/src/analyze/lit-analyzer.ts
@@ -64,6 +64,7 @@ export class LitAnalyzer {
 		} else if (document instanceof HtmlDocument) {
 			return this.litHtmlDocumentAnalyzer.getDefinitionAtOffset(document, offset, request);
 		}
+		return;
 	}
 
 	getQuickInfoAtPosition(file: SourceFile, position: number): LitQuickInfo | undefined {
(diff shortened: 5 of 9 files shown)


runem:
That sounds like an interesting project! Out of curiosity, what exactly are you prototyping? And is there any way I can help you? 😊 

In generally I like stricter type checking better, so I'm fine with "noImplicitReturns". Thanks for contributing 🎉

A side note to be aware of: During this month I plan on changing the structure of this repository a bit. I have been working on separating "typescript plugin" specific code from the analyzing logic on this branch: https://github.com/runem/ts-lit-plugin/tree/lit-analyzer. In addition, perhaps I will turn this repo into a mono-repo with 3 projects: lit-analyzer, ts-lit-plugin and vscode-lit-plugin (just a heads up).

rictic:
> That sounds like an interesting project! Out of curiosity, what exactly are you prototyping? And is there any way I can help you? 😊

Yeah! So, I'm prototyping what it would look like to get your lit-plugin running as part of the dev experience for google's devs. There's two parts of this, running checks as part of the build system, and loading the plugin into the TypeScript compiler for editors. So far I've got something working for both scenarios, but I suspect that I'm doing some reinvention of the wheel.

For bazel, I'm implementing the wrap method of (https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/plugin_api.ts) and overriding getSemanticDiagnostics to return the original program's diagnostics plus the results from litAnalyzer.getDiagnosticsInFile(file).

For hooking into tsserver for non-vscode editors, I wrote a small patch to the internal version tsserver.js (like I said, hacky prototyping! nothing that would actually land) to wrap its createProgram result. Now I'm trying to use the plugins array in tsconfig instead.

So, I'm doing a bit of stumbling around, hooking things together, but I feel like I'm missing important parts of the picture!

So, far I've got two questions I'm still trying to figure out:

1. Is there code to convert a LitDiagnostic to a TypeScript Diagnostic? It seems like there must be, but I can't find it! Wasn't too hard to write my own, but better to reuse code of course.

2. From what I've been able to work out, when you put a plugin into a tsconfig file, it results in tsserver resolving the plugin module name, which should resolve to a commonjs module with a default export that's a PluginModuleFactory (https://github.com/microsoft/TypeScript/blob/4ae091fa9ef3c4043530908d03525aca108fdd54/src/server/project.ts#L100). Is that right? If so, does ts-lit-plugin implement that API anywhere? My understanding is that it does, but I haven't figured out where.

So far I've been super impressed with what I've seen of this code. It implements so many awesome features! And thanks for offering to help!

rictic:
Hm, I think the issue common to both my questions is that master of this repo seems to be for the package lit-analyzer? When I install ts-lit-plugin from npm I see code more along the lines of what I expect.

runem:
That sounds really exciting!

I'm so sorry for the confusion. Yes, the master branch and the published code are out of sync right now.

2 months ago, I started a big refactoring. Midway into this refactor, I chose to clean up/document/add tests to "web-component-analyzer" and "ts-simple-type" before continuing. Therefore, at the moment, many things are out of sync in this repo (as well as vscode-lit-plugin).

1. [This file](https://github.com/runem/vscode-lit-plugin/blob/lit-analyzer/packages/ts-lit-plugin/src/ts-lit-plugin/translate/translate-diagnostics.ts) has code for converting between LitDiagnostic and TypeScript diagnostic. However it sounds like you need an instance of [this class (TsLitPlugin)](https://github.com/runem/vscode-lit-plugin/blob/lit-analyzer/packages/ts-lit-plugin/src/ts-lit-plugin/ts-lit-plugin.ts) which abstracts away the conversion. If you need that class please be aware that as of now I don't publish TS declarations for ts-lit-plugin (and no barrel export), but I will make sure to do that in the future. Right now you can find it in `node_modules/ts-lit-plugin/lib/ts-lit-plugin/ts-lit-plugin.js` if using version 1.0.6.

2. [This file](https://github.com/runem/vscode-lit-plugin/blob/lit-analyzer/packages/ts-lit-plugin/src/index.ts) exports the PluginModuleFactory that you are looking for (node_modules/ts-lit-plugin/lib/index.js)

As you can see the current state of the repo is a bit messy, with a lot of work being done one other branches. This is mainly because the refactoring took longer time than expected, but I will get to clean up/adding tests to this repo very soon. I'm on vacation next week, but the week after, I will make it my first priority to clean up this library. I'll keep you updated 👍 

Until now, I haven't been focusing on a public API for using this library, but after the refactoring I will write up a clear guide on how to use the public API.

Please feel free to write me if you have more questions or if you are experiencing problems. I will gladly help you as much as possible 😊

rictic:
Thanks for the write up!

I grabbed the versions off of npm and was able to make rapid forward progress. I ran it across all the code using lit in google3 and found it to be super fast, gave helpful errors, and was easy to extend. Just super, super awesome.

My plan (if things go as hoped) would be to hook it into the build system. We'd use it in a single configuration for all projects that depend on lit-html or lit-element, and any failures would be considered build errors, requiring that they be fixed.

In general, it was quick and easy to fix the vast majority of errors, and for the rest I'll be sending PRs that add additional options for our use case. Here's some really rough notes I've been taking as I went through the errors:

 * disallowing unknown attributes is probably too strict. would be super useful though…
 * love the idea of the missing import check. however… due to the need to support different ways of importing some code (e.g. goog.module), we have some important code that re-exports elements declared in other files. for the moment, gonna skipMissingImports
 *  CSS parse failures everywhere, seems like the vscode css parser is stricter than we'd like, and people will also compose css tags out of other css. i.e.

```javascript
const size = css`10px`;
const style = css`
  :host {
    ${size}
  }`;
```
* I'm gonna look into adding a `skipCssParserErrors` option
* An interesting issue, for code like: `<button @tap=${handler}">` (note the trailing `"`)
  *  It declares an error as though the type of handler was the empty string literal. This code is wrong (or at least, it's not great), but the error message is a bit confusing. This came up more than once, surprisingly! So apparently it's a typo people make sometimes
* Type checking our security code. Tricky!
    * Internally we use a bit of code that combines https://github.com/Polymer/lit-html/pull/750 and the closure library's safe types, (e.g. https://google.github.io/closure-library/api/goog.html.SafeHtml.html).   
    * I'm working on open sourcing this stuff, but our open source solution might instead be to use https://github.com/WICG/trusted-types which side steps a lot of issues from using closure libraries.
    * In practice what both options mean is that the type of some built in attributes and properties have changed. For example, an iframe's src isn't string, but TrustedResourceUrl.
    * Can't do normal type comparison (in any way that I can see), because it's not clear how to get a reference to the trusted type objects.
    * However, it would be fine to just check that the type names (SafeHtml, SafeStyle, SafeUrl, TrustedResourceUrl, etc) are as expected. It doesn't have to be fully correct, there's runtime security checks.
  * I've got a suuuuuuuper hacky version of this that seems to work ok, but will need some more thought

daKmoR:
that sounds really awesome 🤗 

lot's of great progress 👍  if there is anything we can help with just let us know 🤗

runem:
I’m really happy to hear that it worked for you! :tada:

1. I made the default config strict, because I find it very useful in my own projects and because like that it forces me to document my components. However, do you think it would be better the other way around and have a non-strict default config? That might be more appropriate, especially because there are still some cases where web-component-analyzer fails to find all attrs/props.

2. Regarding your problems with css errors, I’ll look into a fix that makes it possible to compose CSS tags (which by them selves could be just values or singles key-value rules). I also like the idea of splitting the CSS features into multiple fined grained rules like “CSS parser errors”.  Note that (on the lit-analyzer branch), I have already been working on dividing some functionality into fine grained, clearly defined rules, that can be turned off or specified as worning/error.

3. That’s a very interesting issue. I’ll have to fix the issue directly in the HTML parser. You will also encounter the same in code like this: `<input @change=${handler}/>`. 

4. Type checking these security types seems like an interesting challenge. I’m up for the challenge if you would like to see support for closure safe types. I think I would like to add it as an optional feature that can be enabled in the config. Let’s open an issue where we can specify the requirements. Eg is there an requirement for where types like “SafeHtml” can be used, or should it be assignable to anything? Should other types like a plain “string” result in an error? In addition it would be interesting to look into how Typescript plan to handle standard trusted types for HTML elements.  

Thank you so much for your contributions @rictic, - they are greatly appreciated! Be aware that I’m on vacation right now, so I don’t have that much time atm, but Ill make sure to look at all your PR’s/issues when I get home next week :blush:

I’ll make sure to write up a contribution guide/description of the architecture, add tests and create some issues for all of this so that it becomes easier to contribute in the future. And thanks for offering your help @daKmoR :hugs:

PR #24, opened 2019-07-08
Add a config option dontSuggestConfigChanges

rictic:
This one's pretty obscure for most people, so I'm not sure it's even worth documenting in the README. It's useful in our case because in the internal build system at google we want to use a single canonical configuration for the plugin, in order for things to be simpler and more consistent.

As a result, it'd be nice if we could configure the plugin to not suggest making configuration changes, because changes in configuration would be rare and global, so not a reasonable response to any single error.

Also made a handful of minor changes to message wording (to use consistent quote marks, that sort of thing).


runem:
I completely understand the need for this option. I also think, like you, that it's not worth documenting in the README.

PR #26, opened 2019-07-10
Run npm run prettier:write

abdonrd:
Run `npm run prettier:write`


runem:
Thanks for your contribution :tada: Prettier is setup to run for each commit, so I find it strange that some files are not formatted correctly. I'll have to look into this :-)

PR #27, opened 2019-07-11
Support both the parse5@5.0 and the @4.0 location info APIs

rictic:
If supporting both is a maintenance burden, we can solve this
downstream, but so far it seems pretty simple.


runem:
I think it's completely fine to support both 4 & 5 major versions of parse5 👍 Thanks for your contribution!

PR #37, opened 2019-08-11
Add the ability to emit index entries.

rictic:
I'm prototyping a Kythe TypeScript indexer plugin and things are fitting together pretty well! This same method should also be useful for generating the language server index format.

This starts by adding a method on the analyzer that will yield all of the information that the analyzer has determined about a file. I started with element and attribute references as they seem like the most valuable. What else could we emit here?


runem:
That sounds really exciting!

I'm not 100% sure how much Kythe covers, but off the top of my head I can't think of much else to include :-) 

Maybe it would be interesting to see which custom elements are available in a given file? I already analyze this for the `no-missing-import` rule.

justinfagnani:
@rictic do you still want to move this one forward? If you rebase I'll review!

PR #39, opened 2019-08-11
Implement a bazel plugin

rictic:
This implements a bazel diagnostic plugin, for integrating ts-lit-plugin checks into the bazel ts_library implementation.

Once this is released I'll start work on open sourcing the bazel side of this integration. The goal would be that that a bazel user could switch on the ts-lit-plugin simply by setting a parameter (like `use_lit_plugin = True`) on their ts_library rule.


runem:
That sounds really exciting, thanks for the contribution! :tada:

runem:
Version `1.1.8`, which includes the Bazel plugin, has just been released :+1:

PR #47, opened 2019-08-18
wip: rules modularised

### LINKED ISSUE #46: Organising rules into individual modules
As part of giving this analyzer a go, i've been reading through the sources and what not.

One thing you could do IMO to tidy things up a lot and make the code base much easier to follow is introduce rules as their own concepts/modules.

What I mean is:

* Move all "rules" into their own file as functions which follow a new `RuleModule` type and are used like factories to create a `RuleVisitor` or some such thing
* Move all "reports" to the analyzer context
* Refactor the "analyzers" (e.g. `LitHtmlDocumentAnalyzer`) to instead load an array of rules and visit each node via them

I've already done most of this in a branch to see what would happen. So if you agree its a fair idea, ill push a PR up at some point when i get it all out of my head.

43081j:
Resolves #46.

Basically moved all rules into individual files, stripped down all the document analysis so it just uses "visitors" rather than caring which ones, and shuffled a bunch of things here and there.

## WIP

So this is the kinda idea i had, i described it already in #46 but do take a look at the code just to see if its something you'd be interested in seeing me finish.

Its just something that would help keep the codebase clean IMO especially as it grows.

@rictic if you have any thoughts too LMK as this is mostly based on what i learnt from work i did on the polymer analyzer with you


43081j:
_it builds!_

which is either impressive or worrying.. 🙈

runem:
Like mentioned [here](https://github.com/runem/lit-analyzer/pull/50) 1.2.0 has all the functionality discussed here. Therefore I'm closing this PR now. Please open another PR/issue if you still think there are some things we could improve on :+1:

PR #50, opened 2019-08-22
Rule iteration & short-circuiting

43081j:
again WIP but what about something like this?

we don't need the cruft of the diagnostics classes anymore at this point so it all got pulled up into the analyzer.

at the min there's no implementation to maintain the map because we don't have a `Node` type.

what do you suggest for that? Ideally we just hold around a `Set<Node>` where a `Node` is a html or css node. Maybe we can just introduce a type like i have in the default context?


packages/lit-analyzer/src/analyze/lit-analyzer-context.ts
```diff
@@ -21,9 +25,12 @@ export interface LitAnalyzerContext {
 	readonly definitionStore: AnalyzerDefinitionStore;
 	readonly logger: LitAnalyzerLogger;
 	readonly rules: RuleModule[];
+	readonly reports: LitDiagnostic[];
 	updateConfig(config: LitAnalyzerConfig): void;
 	updateDependencies(file: SourceFile): void;
 	updateComponents(file: SourceFile): void;
+	report(data: LitDiagnostic): void;
+	hasReports(node: HtmlNode | HtmlNodeAttr | HtmlNodeAttrAssignment): boolean;
```
runem: I think `report`, `hasReports` and `reports` would be better on `LitAnalyzerRequest` (which will soon be renamed to context). The existing `LitAnalyzerContext` (which will soon be renamed to base context) is the same instance during the life time of lit-analyzer but the `LitAnalyzerRequest` is created per call to the service in the method `makeRequest`: https://github.com/runem/lit-analyzer/blob/5f4b3935ed99c99144a0f2215d32a8ec43ed5397/packages/lit-analyzer/src/analyze/lit-analyzer.ts#L293

If we keep it on `LitAnalyzerContext` we would have to reset `_reportedNodes` and `reports` after the `getDiagnostics` method call so that they are ready for the next request. If they are on the `LitAnalyzerRequest` they would just be thrown away afterwards, because `makeRequest` creates a new instance each time.

43081j: yep it should be on the request, my mistake.

the `makeRequest` seems a little odd to me when i keep bumping into it too. as it takes an instance and returns a plain object with the same properties. maybe it would be better to new up a new context? or have a teardown/clear method?

runem: Yeah I also think `makeRequest` is bit awkward method right now. I think it's a strength that we are able to construct a temporary instance per request, so I think we should continue having these two data structures.

A possible solution could be to:

 1. Move `makeRequest` method to `LitAnalyzerContext`. When we want to new up a request we could call `this.context.makeRequest()` instead.
 2. Refactor `LitAnalyzerRequest` so that it's not extending `LitAnalyzerContext` but instead has a `context` property (or `baseContext` later on) which is of type `LitAnalyzerContext`.
 3. Make a class `class DefaultLitAnalyzerRequest implements LitAnalyzerRequest` which takes the `LitAnalyzerContext` in the constructor and sets it as `context` property. This class could also implement `report`, `hasReports` and `reports`.

43081j: if we want report and hasReports on the request, it really should be a class tbh.

IMO a small class which wraps the base (i.e. has an internal `analyzerContext` or something) would be fine to replace the request.

runem: I agree that it should be a class that wraps a base context. That's also what I'm suggesting in the possible solution above (see step 3). I just changed "continue having these two data structures" to "continue having **two separate** data structures" to make it more clear that I'm suggesting changing it to a class instead of an object :-)

43081j: i tried, it becomes a little weird. we end up with a context (the request) which contains a context..

we also dont want to have to do what we do now (clone it each time).

so im not too sure what the best approach is, other than having `context.context` 😂 where the first is the request

43081j:
@runem it would still be super nice and a good clean up if we could pull all the rule iteration up into a single point of truth IMO.

this would mean we can drop most or all of the 'diagnostics' classes/overhead and have a single point where we check which rules are enabled etc.

is that something you're going to look into? happy to do a new branch off master (or `rules`) to play around with the idea or rebase this one and improve it.

runem:
`1.2.0` has all the functionality discussed here. `context` and `request` have also been merged into one. Therefore I'll close this PR now (I hope it's okay) :-) Please open another PR/issue if you still think there are some things we could improve on.

PR #62, opened 2019-11-06
Initial security type checking.

### LINKED ISSUE #28: Type checking with security sanitization in place
When running in an environment with https://github.com/WICG/trusted-types* the types of some security sensitive built-in attributes and properties are effectively changed. For example, an iframe's src isn't string, but TrustedUrl.

Full list of overrides here: https://wicg.github.io/trusted-types/dist/spec/#enforcement-in-sinks

As the spec isn't yet finalized, the types aren't yet included automatically in the main typings. Further, inside google we use a slightly different system for the time being based on closure library types (e.g. https://google.github.io/closure-library/api/goog.html.SafeUrl.html).

Type typing these with these types doesn't have to be perfect, the canonical security layer is at runtime.

*or when running lit-html with https://github.com/Polymer/lit-html/pull/750 enabled, which is kinda-sorta a ponyfill of the trusted types spec purely for lit-html templates

rictic:
This adds basic support for type checking code with Safe Types sanitization in place. This can be trivially extended to support checking with Trusted Types, but I'm starting simple.

Trusted Types is also still working through the standards process, so I don't want to ship anything prematurely here. See the conversation at https://groups.google.com/a/chromium.org/g/blink-dev/c/Il-wfnw9TAw

Fixes #28

### DIFF
--- a/packages/lit-analyzer/src/analyze/lit-analyzer-config.ts
+++ b/packages/lit-analyzer/src/analyze/lit-analyzer-config.ts
@@ -136,9 +136,12 @@ export function litDiagnosticRuleSeverity(config: LitAnalyzerConfig, ruleName: L
 
 export type LitAnalyzerLogging = "off" | "error" | "warn" | "debug" | "verbose";
 
+export type LitSecuritySystem = "off" | "ClosureSafeTypes";
+
 export interface LitAnalyzerConfig {
 	strict: boolean;
 	rules: LitAnalyzerRules;
+	securitySystem: LitSecuritySystem;
 
 	disable: boolean;
 	logging: LitAnalyzerLogging;
@@ -165,6 +168,7 @@ export function makeConfig(userOptions: Partial<LitAnalyzerConfig> = {}): LitAna
 	return {
 		strict: userOptions.strict || false,
 		rules: makeRules(userOptions),
+		securitySystem: userOptions.securitySystem || "off",
 
 		disable: userOptions.disable || false,
 		logging: userOptions.logging || "off",
--- a/packages/lit-analyzer/src/analyze/lit-analyzer-config.ts
+++ b/packages/lit-analyzer/src/analyze/lit-analyzer-config.ts
@@ -160,11 +160,27 @@ export interface LitAnalyzerConfig {
 	customHtmlData: (string | HtmlData)[] | string | HtmlData;
 }
 
+function expectNever(never: never) {
+	return never;
+}
+
 /**
  * Parses a partial user configuration and returns a full options object with defaults.
  * @param userOptions
  */
 export function makeConfig(userOptions: Partial<LitAnalyzerConfig> = {}): LitAnalyzerConfig {
+	let securitySystem = userOptions.securitySystem || "off";
+	switch (securitySystem) {
+		case "off":
+		case "ClosureSafeTypes":
+			break; // legal values
+		default:
+			// Log an error here? Or maybe throw?
+			expectNever(securitySystem);
+			// Unknown values get converted to "off".
+			securitySystem = "off";
+	}
+
 	return {
 		strict: userOptions.strict || false,
 		rules: makeRules(userOptions),
--- a/packages/lit-analyzer/src/analyze/util/type/is-assignable-binding-under-security-system.ts
+++ b/packages/lit-analyzer/src/analyze/util/type/is-assignable-binding-under-security-system.ts
@@ -3,6 +3,7 @@ import { SimpleType, toTypeString, SimpleTypeKind } from "ts-simple-type";
 import { LitAnalyzerRequest } from "../../lit-analyzer-context";
 import { LitHtmlDiagnostic, LitHtmlDiagnosticKind } from "../../types/lit-diagnostic";
 import { LitAnalyzerRuleName } from "../../lit-analyzer-config";
+import { isLitDirective } from "../directive/is-lit-directive";
 
 /**
  * If the user's security policy overrides normal type checking for this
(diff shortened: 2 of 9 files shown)


rictic:
There's something slightly odd going on, at least when I run this locally. The analyzer doesn't seem to know about any properties on the `<script>` element. It knows about attributes, but not properties. This is a regression (and very possibly a result of my local environment being in a weird state). I haven't yet tracked down what's going on in any detail.

rictic:
Also may depend on https://github.com/runem/ts-simple-type/pull/56

packages/lit-analyzer/src/analyze/lit-analyzer-config.ts
```diff
@@ -165,6 +168,7 @@ export function makeConfig(userOptions: Partial<LitAnalyzerConfig> = {}): LitAna
 	return {
 		strict: userOptions.strict || false,
 		rules: makeRules(userOptions),
+		securitySystem: userOptions.securitySystem || "off",
```
aomarks: Does this come straight from a user config file, or can we assume it's validated as a partial config already? (Otherwise should this validate and throw?)

packages/lit-analyzer/src/analyze/util/type/is-assignable-binding-under-security-system.ts
```diff
@@ -0,0 +1,109 @@
...
+	if (!matchesAtLeastOneNominalType(overriddenTypes, typeB)) {
+		const nominalType: SimpleType = {
+			kind: SimpleTypeKind.INTERFACE,
+			members: [],
+			name: "A security type"
+		};
+		return [
+			{
+				kind: LitHtmlDiagnosticKind.INVALID_ATTRIBUTE_EXPRESSION_TYPE,
+				message: `Type '${toTypeString(typeB)}' is not assignable to '${overriddenTypes.join(" | ")}'.`,
```
aomarks: Would it be helpful to add something like "under ${securityPolicy} security policy", so that users get a hint about where to look to understand why they couldn't assign?

packages/lit-analyzer/src/analyze/util/type/is-assignable-in-attribute-binding.ts
```diff
@@ -31,6 +32,21 @@ export function isAssignableInAttributeBinding(
...
+			// Purely static attributes are never security checked, they're handled
+			// in the lit-html internals as trusted by default, because they can
+			// not contain untrusted data, they were written by the developer.
+			//
+			// For everything else, we may need to apply a different type comparison
+			// for some security-sensitive built in attributes and properties (like
+			// <script src>).
+			const securityDiagnostics = isAssignableBindingUnderSecuritySystem(htmlAttr, { typeA, typeB }, request, "no-incompatible-type-binding");
+			if (securityDiagnostics !== undefined) {
+				// The security diagnostics are binding. Note that this may be an
```
aomarks: Should this be "for the binding", or am I missing another meaning of "are binding"?

packages/lit-analyzer/src/analyze/lit-analyzer-config.ts
rictic: Following convention here, I didn't throw. Added some validation, since I didn't see other validation. @runem what do you think?

packages/lit-analyzer/src/analyze/util/type/is-assignable-binding-under-security-system.ts
rictic: +1, done

packages/lit-analyzer/src/analyze/util/type/is-assignable-in-attribute-binding.ts
rictic: Good point, using the word "binding" here is confusing because of the more salient other meaning. Clarified

packages/lit-analyzer/src/analyze/lit-analyzer-config.ts
runem: The configuration given to `lit-analyzer` is not validated as of now. First of all, I think the CLI should throw on invalid values, but `ts-lit-plugin` should only log warnings (no reason to crash when running in the IDE). 

Therefore, I think that we should implement one of these solutions:
a. export a `validateConfig` function that validates the raw, partial config.
b. extend `makeConfig` with an option to assert values (perhaps a callback that is called on invalid config values).

I also think `makeConfig` should properly fall back to default values when encountering an invalid value (which it doesn't do for all attributes right now). 

I created an issue for it here https://github.com/runem/lit-analyzer/issues/65 so that it can be done in a future, separate PR 👍

packages/lit-analyzer/src/analyze/util/type/is-assignable-binding-under-security-system.ts
```diff
@@ -0,0 +1,139 @@
...
+	}
+	const [argType] = type.argTypes;
+	if (argType.type.name !== "Part" && argType.type.name !== "NodePart") {
+		return false;
+	}
+	if (type.returnType === undefined || type.returnType.kind !== SimpleTypeKind.VOID) {
+		return false;
+	}
+	return true;
+}
```
runem: Is there a reason you didn't use `isLitDirective` from https://github.com/runem/lit-analyzer/blob/master/packages/lit-analyzer/src/analyze/util/directive/is-lit-directive.ts? Feel free to extend/refactor that function if needed :-)

runem: @rictic, any updates here? If you don't have time to combine the two functions, I'll merge this PR and combine them myself :blush: :+1:

rictic: Yeah, just had to take a bit to get back to this. isLitDirective is more complete and careful, +1 to reusing it!

PR #63, opened 2019-11-08
Use < and > in tag name error messages.

rictic:
IMO it's faster to read and know that it's a tag name if it's written <like-this>


runem:
Thanks :tada:

PR #64, opened 2019-11-08
Transpile only in ava config

rictic:
This gives faster test running, though it does not validate that the code type checks. `npm run build` will, and when developing your editor will tell you about type errors.


rictic:
This is just a preference thing, but IMO this is nicer.

runem:
I agree, thanks! ☀️

PR #83, opened 2020-02-22
Handle circular type refs in matchesAtLeastOneNominalType

rictic:
Using the simple types cache more aggressively resulted in one compilation failing due to a circular type ref here.


runem:
As mentioned here (https://github.com/runem/lit-analyzer/pull/102#issuecomment-653777958) `ts-simple-type` now supports caching of recursive types, and `CIRCULAR_TYPE_REF ` has been removed, so there is no longer need for this check. Therefore I'll close this PR, but thanks for the contribution nevertheless :-)

PR #88, opened 2020-03-15
Closure security needs a safe type for <source src>

rictic:


runem:
Looks great, thanks :-)

PR #91, opened 2020-04-16
Handle more cases in isLitDirective

rictic:
There's now a BooleanAttributePart, and a directive that works on
multiple parts can take a union of Part types.

### DIFF
--- a/packages/lit-analyzer/src/analyze/util/directive/is-lit-directive.ts
+++ b/packages/lit-analyzer/src/analyze/util/directive/is-lit-directive.ts
@@ -1,5 +1,7 @@
 import { SimpleType, SimpleTypeKind } from "ts-simple-type";
 
+const partTypeNames: ReadonlySet<string | undefined> = new Set(["Part", "NodePart", "AttributePart", "BooleanAttributePart", "PropertyPart"]);
+
 /**
  * Checks whether a type is a lit directive.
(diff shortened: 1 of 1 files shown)


packages/lit-analyzer/src/analyze/util/directive/is-lit-directive.ts
```diff
@@ -1,5 +1,7 @@
 import { SimpleType, SimpleTypeKind } from "ts-simple-type";
 
+const partTypeNames: ReadonlySet<string | undefined> = new Set(["Part", "NodePart", "AttributePart", "BooleanAttributePart", "PropertyPart"]);
```
justinfagnani, line 3: Either add `EventPart` to this list, or instead check if the argument type is assignable to `Part`.

runem:
This looks great! @rictic I'll merge this PR as soon as [the comment](https://github.com/runem/lit-analyzer/pull/91#discussion_r416277692) has been resolved :-)

I also think we should add `live`, `templateContent` and `unsafeSVG` to the `no-invalid-directive-binding` rule, in a separate PR, that checks if built-in directives are used in correct bindings. I've created an issue [here](https://github.com/runem/lit-analyzer/issues/99).

runem:
I'll merge this PR tomorrow and add `EventPart` to `partTypeNames` :-)

PR #93, opened 2020-04-27
performance: import traversal

43081j:
analysis takes an incredible amount of time in some of my repos, so i have profiled it and tested it today...

almost all the CPU time is lost to traversal of the dependency graph, i.e. simply the massive amount of recursive calls in `visitDependencies` reach an unreasonable level.

what do you think of this?

the change is to **only visit imported modules which we imported directly, not from external libraries**.

im making these assumptions:

* web-analyzer has already parsed direct imports by this point i imagine, so we already know what elements (and their properties etc) a direct import provides
* it is bad practice to rely on deeply nested imports out of your control to expose an element, you should import it directly

given these two assumptions, all imported elements must be direct imports.

Quick timings in one of my repos:

Before:

```bash
real 2m6.236s
user 2m11.438s
sys 0m3.328s
```

after:

```bash
real 0m13.535s
user 0m18.063s
sys 0m2.031s
```

Same number of errors and warnings came back in all the repos i tried.

cc @runem

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[*]
+indent_size = 1
+indent_style = tab
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,7 @@
 root = true
 
 [*]
-indent_size = 1
+charset = utf-8
+indent_size = tab
 indent_style = tab
+tab_width = 2
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 /.vsCode
 .idea
+.editorconfig
 node_modules
 *.txt
 *.tgz
--- a/packages/lit-analyzer/src/analyze/parse/parse-dependencies/visit-dependencies.ts
+++ b/packages/lit-analyzer/src/analyze/parse/parse-dependencies/visit-dependencies.ts
@@ -56,7 +56,9 @@ export function visitDependencies(node: Node, context: IVisitDependenciesContext
 			return;
 		}
 	} else if (context.ts.isImportDeclaration(node) || context.ts.isExportDeclaration(node)) {
-		if (node.moduleSpecifier != null && context.ts.isStringLiteral(node.moduleSpecifier)) {
+		if (node.moduleSpecifier != null && context.ts.isStringLiteral(node.moduleSpecifier) &&
+				context.ts.isSourceFile(node.parent) &&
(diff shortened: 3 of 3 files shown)


runem:
Wow, great catch!

One concern I have with this change is that it's no longer possible to depend on barrel modules from a library. I agree that it's bad practice to rely on deeply nested imports from an external library, but barrel modules can be used as a controlled and useful way of exposing elements.

Here's and example of how we do it in my company:

**./src/my-element.ts**
```typescript
import "@ideanote/atoms/lib/button";
```

**./node_modules/@ideanote/atoms/lib/button/index.d.ts**
```typescript
export * from "./at-button";
```

**./node_modules/@ideanote/atoms/lib/button/at-button.d.ts**
```typescript
...

declare global {
    interface HTMLElementTagNameMap {
        "at-button": AtButton;
    }
}
```

This change would have us to change all of the library imports to this instead:

```typescript
import "@ideanote/atoms/lib/button/at-button";
```

If possible I would really love to find a performant solution where this kind of pattern become possible. For example by introducing caching, only following external exports X levels deep, or only following the chain of exports in external modules consisting purely of exports :-)

43081j:
Simply changing this to use a max depth would solve your problem there and defaulting to 2 levels deep (i.e. direct imports and their direct imports).

I will update it if that works for you

43081j:
something like what i just pushed but maybe configurable some place

runem:
I think it looks like a fine solution :blush: I still haven't built the functionality to configure rules, but I expect it will be the same as [eslint](https://eslint.org/docs/user-guide/configuring#using-configuration-files). 

One thing: the newly added `.editorconfig`-file crashes `prettier` because it tries to reformat it on `npm run prettier:write`, so we would either have to add an entry to `.prettierignore` or `.gitignore`. If you want to keep the file I would love it to look something like this:

```
root = true

[*]
charset = utf-8
indent_size = 2
indent_style = tab
```

43081j:
it'd be good to keep it since it means there's no risk of people accidentally introducing mixed formatting.

the sources i've seen so far have been 1 tab deep, too, are you sure indent_size should be 2?

runem:
Sorry, I was confusing `indent_size` with `tab_width`. I would prefer a `tab_width` of 2 :-)

43081j:
there you go i updated it

runem:
Great, thanks! I'll merge now

Issue #92, opened 2020-04-27
Feature Request: warn for unknown attribute that ends with '?'


rictic:
This is a common mistake Polymer users make when they're moving to lit, where instead of doing:

```html
  <div ?hidden=${false}>
```

They do

```html
  <div hidden?=${false}>
```

To lit-analyzer I think this looks like a binding to the unknown attribute `'hidden?'` but the intent is fairly clear.

PR #95, opened 2020-04-28
no-legacy-attribute rule

43081j:
could this be a solution to #92 in a way?

i haven't added docs and finished it up just in case its not the kind of solution we're looking for.

it just detects legacy polymer binding syntax (`foo$=`, `foo?=`, `foo="[[bar]]"`, `foo="{{bar}}`).

cc @rictic @runem


runem:
This looks great, thanks for the contribution! I will merge as soon as the rule is described in the readme :+1: 

The description of the rule would have to be added to [docs/readme/rules.md](https://github.com/runem/lit-analyzer/blob/master/docs/readme/rules.md). You don't have to run `npm run readme`, because it seems like the formatting of `packages/*/README.md` is out of sync. I'll take a look at fixing that after the merge :-)

rictic:
This is really great!

43081j:
here you go i rebased on master and updated docs

runem:
Fantastic, thanks! ☀️ I'll merge now :-)

PR #86, opened 2020-03-10
Fixes typo about event type detection

lourd:
This is just a one word change. I think all of the other lines being shown as changed have something to do with line endings or whitespace.

PR #96, opened 2020-04-29
Set maxWarnings as -1 to avoid fail CLI when just warnings

### LINKED ISSUE #94: The warnings should not fail in the CLI
I expect the warnings should not fail in the CLI, just the errors.

Screenshot 2020-04-28 at 20 03 09

```
> lit-analyzer "src/{components,pages}/**/*.ts" --strict

Analyzing 4 files...

./src/components/app-shell.ts

    Unknown tag <pwa-install-button>.
    67:  <pwa-install-button>
    no-unknown-tag-name

    Unknown tag <pwa-update-available>.
    71:  <pwa-update-available>
    no-unknown-tag-name

  ✖ 2 problems in 1 file (0 errors, 2 warnings)

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! pwa-starter@0.1.0 lit:check-types: `lit-analyzer "src/{components,pages}/**/*.ts" --strict`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the pwa-starter@0.1.0 lit:check-types script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

An example with ESLint:

Screenshot 2020-04-28 at 20 06 32

abdonrd:
First I run the `npm run prettier:write` command https://github.com/runem/lit-analyzer/commit/89d691f8b669d1b3bf2d9fc339207296cad6b1e9

Fix #94

### DIFF
--- a/packages/lit-analyzer/src/cli/analyze-command.ts
+++ b/packages/lit-analyzer/src/cli/analyze-command.ts
@@ -111,7 +111,11 @@ export async function analyzeCommand(globs: string[], config: LitAnalyzerCliConf
 		printText(statsText, config);
 	}
 
-	return stats.diagnostics <= (config.maxWarnings || 0) || config.maxWarnings === -1;
+	if (stats.errors !== 0) {
+		return false;
+	}
+
+	return stats.warnings <= (config.maxWarnings || 0) || config.maxWarnings === -1;
 }
 
 function getFormatter(format: FormatterFormat): DiagnosticFormatter {
--- a/packages/lit-analyzer/README.md
+++ b/packages/lit-analyzer/README.md
@@ -38,7 +38,7 @@ npm install lit-analyzer -g
 lit-analyzer src
 lit-analyzer "src/**/*.{js,ts}"
 lit-analyzer my-element.js
-lit-analyzer --format markdown --outFile result.md 
+lit-analyzer --format markdown --outFile result.md
 ```
 
 [![-----------------------------------------------------](https://raw.githubusercontent.com/andreasbm/readme/master/assets/lines/rainbow.png)](#configuration)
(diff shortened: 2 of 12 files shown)


packages/lit-analyzer/src/cli/analyze-command.ts
```diff
@@ -111,7 +111,11 @@ export async function analyzeCommand(globs: string[], config: LitAnalyzerCliConf
 		printText(statsText, config);
 	}
 
-	return stats.diagnostics <= (config.maxWarnings || 0) || config.maxWarnings === -1;
+	if (stats.errors !== 0) {
```
43081j, line 114: you'll probably also want to update the `failFast` condition further up to only fail if we have errors or have set `maxWarnings` to less than how many warnings were seen so far

abdonrd: As the documentation says:

>Exit the process right after the first problem has been found

I think the current behavior is okey.

packages/lit-analyzer/src/cli/analyze-command.ts
```diff
@@ -111,7 +111,11 @@ export async function analyzeCommand(globs: string[], config: LitAnalyzerCliConf
...
+		return false;
+	}
+
+	return stats.warnings <= (config.maxWarnings || 0) || config.maxWarnings === -1;
```
abdonrd, line 118:
```suggestion
	return stats.warnings <= config.maxWarnings! || config.maxWarnings === -1;
```

@runem I don't know if you prefer the `|| 0` for some reason.
If not, you can accept this proposal. :)

abdonrd: @43081j got it, thanks!

packages/lit-analyzer/src/cli/analyze-command.ts
runem: I agree with @43081j, - I'll implement that after merging this PR. The word `problem` in that context was meant to be interpreted as "the first diagnostic that results in a failure", so I think it would make sense to take `--maxWarnings` into account for the `--failFast` option.

runem:
Thank you so much for this PR! I'll go ahead and merge :-)

abdonrd:
Yay! Thanks @runem! I look forward to a new release! 😀

abdonrd:
Hi @runem! When we can expect a new version with this? Thanks in advance!

runem:
I just made this change log by looking at the commit log. These are the changes that will go live with the release of `1.1.11`. These changes seem relativ safe to publish. I'll run the final tests and will to make a publish during this week :blush: 

* Better performance when using the `no-missing-imports` rule. `lit-analyzer` will still check imported modules in your project as usual, however, it will only follow imports 2 levels deep into any imported module from an external dependency (see [#93](https://github.com/runem/lit-analyzer/pull/93))

* New rule: `no-legacy-attribute` which is disabled as default. A common mistake when dealing with Lit in particular is to use the legacy Polymer syntax as seen in earlier versions of Polymer (the predecessor of Lit). This rule catches this mistake (see [#95](https://github.com/runem/lit-analyzer/pull/95))

* The CLI-option `maxWarnings` defaults to `-1` to avoid failing when the analysis found only warnings (see [#96](https://github.com/runem/lit-analyzer/pull/96))

* Added closure security safe type for `<source src>` (see [#88](https://github.com/runem/lit-analyzer/pull/88))

* Fixed typo about event type detection (see [#86](https://github.com/runem/lit-analyzer/pull/86))

abdonrd:
Great, thanks @runem!

runem:
@abdonrd Version 1.1.11 has been published now :-)

abdonrd:
Thanks @runem! 👏

PR #98, opened 2020-04-29
bump dependencies & simplify enabled rules logic

43081j:
do it only in one place and bump some deps.

seems to work fine, let me know if anything looks off.

typescript can't be updated until WCA has the same version bump.

edit: no clue why ci fails, exact same stuff passes locally with a clean  repo (git clean -xdf)

### DIFF
--- a/packages/lit-analyzer/package.json
+++ b/packages/lit-analyzer/package.json
@@ -43,8 +43,8 @@
 		"fast-glob": "^2.2.6",
 		"parse5": "5.1.0",
 		"ts-simple-type": "~0.3.6",
-		"vscode-css-languageservice": "4.0.2-next.1",
-		"vscode-html-languageservice": "2.1.12",
+		"vscode-css-languageservice": "4.1.2",
+		"vscode-html-languageservice": "3.0.3",
 		"web-component-analyzer": "~0.1.17"
 	},
 	"devDependencies": {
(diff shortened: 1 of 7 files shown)


43081j:
closing because 1.2.0 shuffles a whole load of this logic around

runem:
I was actually just writing a comment in here which I will post/add nevertheless:

The reason the CI fails is because of problems with [vsce](https://github.com/microsoft/vscode-vsce) and mono-repositories. Unfortunately, vsce doesn't yet support mono-repositories with linked dependencies. You can read more about the issue here: https://github.com/Microsoft/vscode-vsce/issues/203. In order to get around this problem and debug locally developed dependencies (`ts-lit-plugin` and `lit-analyzer`) in the vscode extension project, I had to make a small script that copies locally built lib files to `packages/vscode-lit-plugin/node_modules/{ts-lit-plugin, lit-analyzer}` (see [copylink script](https://github.com/runem/lit-analyzer/blob/master/scripts.js)). The problem we experience in this PR occurs because `lit-analyzer` has updated dependencies, but the dependencies in `packages/vscode-lit-plugin` are installed using dependencies of the published version of lit-analyzer (thus installing 'old' dependencies). `vsce` checks for out-of-sync dependencies when packaging and crashes. I found a way around this on the 1.2.0 branch, and I can cherry-pick the fix if necessary.

runem:
In general if we change code around rule logic I would love it to be based on the 1.2.0 branch, so we avoid too many merge conflicts :+1: But if you think this (or another PR) is very important for an upcoming release of 1.1.x I'm also fine with handling the merge conflicts :-)

43081j:
its ok thats a shame about the vscode issue though.

ill probably throw together a different branch on top of the 1.2 branch, if anything. i think your new rule sets achieve what i was going for mostly. the dependency updates are nice to have though so i might still do that, as the newer ava has debugging built in (like node's `--inspect`).

Issue #79, opened 2020-01-22
Feature request: a check for closure+tsickle renaming safety


rictic:
Short version: I'd like to implement a rule, disabled by default, that requires that properties declared with `@property` must be public, and that properties declared with the new `@internalProperty` must be private or protected. If the visibility does not match up with the decorator, I'd like to recommend that the code author use the decorator that matches up with their preferred visibility.

Longer version on the what and why:
So at Google our TypeScript passes through tsickle to be come JavaScript, and then is compiled by Closure Compiler. Closure Compiler implements some really powerful optimizations like whole-program dead code elimination (not just tree shaking), but many of these optimizations make assumptions about the code. One of the most invasive such assumptions is property renaming, which will actually rename the fields on a class and everywhere that it's used. For custom elements, we don't want public fields to be renamed, because they could be used inside of raw HTML, in templates, etc, but it's ok if private or protected fields are renamed, because they shouldn't be used from HTML.

With LitElement, any field used with `@property` should not be renamed (after going through tsickle & closure compiler), but a field with `@internalProperty` may be. So we want to recommend the appropriate decorator for a given visibility. (We initially tried moving this complexity into tsickle, but the tsickle maintainers didn't want to make their tool any more complicated, so we've ended up with two decorators).

rictic:
Any advice on implementing this? It looks like in the current architecture rules only apply to HTML, and most of the analysis of the declaration of a custom element takes place in webcomponent-analyzer. I could add the necessary information to webcomponent-analyzer and then add a check in lit-analyzer, but it would mean creating an additional kind of RuleModule.

Is this the architecture you'd recommend?

runem:
I just finished adding support for `@internalProperty` in wca ([this commit](https://github.com/runem/web-component-analyzer/commit/d8809c255c5f19b3ba63eafbbec7f0af5791a636)) :-)

In the past, it was the responsibility of wca to emit diagnostics for decorators/members, but I'm in the process of migrating these into lit-analyzer as rules. In this process, I'm adding a new kind of RuleModule that emits diagnostics for members. So I agree with the architecture that you suggest.

I'll let you know when I'm finished migrating the existing rules from wca to lit-analyzer using the new RuleType, - then I think it would be easy to implement the rule that checks visibility-modifiers + property/internalProperty :+1:

rictic:
Awesome, great to hear!

PR #100, opened 2020-05-09
rule: property visibility mismatch

43081j:
this tries to do something along the lines of what #79 wanted i guess.

this bit is _questionable_:

```ts
	return member.kind === 'property' && member.meta?.attribute === false;
```

to detect if `@internalProperty` was used. it is upto you if you want to introduce something in WCA so component members have an explicit flag when they use that decorator. this should work fine though since making a property not use attributes is close enough to being internal anyway.

### DIFF
--- a/docs/readme/rules.md
+++ b/docs/readme/rules.md
@@ -500,6 +500,25 @@ class MyElement extends LitElement {
 customElements.define("correct-element-name", MyElement);
 ```
 
+#### no-property-visibility-mismatch
+
+When using the `@property` decorator, your property should be publicly visible,
+expected to be exposed to consumers of the element. Private and protected
+properties however, should make use of the `@internalProperty` decorator
+instead.
+
+This rule will ensure public properties use `@property` and non-public
+properties use `@internalProperty`.
+
+The following example is considered a warning:
+
+<!-- prettier-ignore -->
+```ts
+class MyElement extends LitElement {
+	@property() private myProperty: string;
+}
+```
+
 ### Validating CSS
 
 `lit-analyzer` uses [vscode-css-languageservice](https://github.com/Microsoft/vscode-css-languageservice) to validate CSS.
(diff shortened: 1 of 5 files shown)


runem:
You should be able to access the decorator through `member.meta` which is of the following type:

```typescript
interface LitElementPropertyConfig {
    type?: SimpleType | string;
    attribute?: string | boolean;
    node?: {
        type?: Node;
        attribute?: Node;
        decorator?: CallExpression;
    };
    hasConverter?: boolean;
    default?: unknown;
    reflect?: boolean;
}
```

So you would be able to check for `member.meta?.node?.decorator?.identifier.text` to get the name of the decorator :+1:

43081j:
i updated it, didn't realise you could access the typescript node there.

remember this is based on 1.2.0 too FYI

runem:
`member.node` is the node where the property/attribute was found/emitted. This could be a `PropertyDeclaration`, `ExpressionStatement`, `PropertyAssignment`, eg. depending on how it was found.

`member.meta` contains specific metadata of the analysis of the property/attribute. This data-structure right now only contains lit-specific information, but I plan that this data-structure in the future could depend on the specific `WCA`-flavor that emitted it. For example, if there was to be created a Stencil-specific `WCA`-flavor, the meta-field would contain `Stencil`-specific data.  

`member.meta.node.decorator` will always point to the decorator that contains `LitElement`-specific metadata which was found during the analysis of the `PropertyDeclaration`. The decorator would always be either a `property`- or `internalProperty`-`CallExpression` and other decorators would be ignored. If there are multiple decorators, this would still be either the `property` or `internalProperty` decorator. The loop over `member.node.decorators` has already been done in `WCA` and the decorator of interrest would be assigned to `member.meta.node.decorator`, so you can trust the `text` property of that node to be `internalProperty` when that's the case.

43081j:
i updated it. just wasn't aware what the meta node was so thanks for the explanation. i do wonder what'll happen when one day multiple decorators might become a thing in lit, though. maybe it should be an array rather than assuming there will always be just one "useful one" per property.

PR #102, opened 2020-05-13
perf: pass a cache when converting types

43081j:
Ignoring the typescript compilation, i got these times:

1.2.0 - 8 seconds
perf-type-cache - 6 seconds

Just by having this weak map in place. Is this fine?

cc @runem


runem:
[branch 1.2.0](https://github.com/runem/lit-analyzer/pull/112) now depends on v1.0.0 of `ts-simple-type` which comes with lazy types and fixed caching. It's possible to cache types properly now because of the removal of the `CIRCULAR_REF` type. All types are cached as default and there is no need for providing our own cache when interacting with `ts-simple-type` :-) Therefore I'll close this PR now.

PR #107, opened 2020-06-16
Add "autocapitalize" global attribute

dfreedm:
Add the global attribute [autocapitalize](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize), which has possible values of `off`, `none`, `on`, `sentences`, `words`, and `characters`


runem:
Thank you very much for your contribution! I'll merge now. You can expect this fix to be released with version 1.2.0 :-)

PR #126, opened 2020-07-25
no-complex-attribute-binding: take converters into consideration

### LINKED ISSUE #123: lit-plugin(no-complex-attribute-binding) when property has a converter
Hello!

My intended API is
```js
// either
html`<trigger-mutation refetch-queries="OneQuery,AnotherQuery"></trigger-mutation>`
// or
html`<trigger-mutation .refetchQueries="${['OneQuery', 'AnotherQuery']}"></trigger-mutation>`
```

As such, I've defined `refetchQueries` as:
```ts
class TriggerMutation extends ApolloMutation {
  @property({ attribute: 'refetch-queries', converter: {
    fromAttribute(string: string): string[] {
      return !string ? undefined : string.split(',').map(x => x.trim()).filter(Boolean);
    },
    toAttribute(value: string[]): string {
      return Array.isArray(value) ? value.join(',') : null;
    },
  } })
  refetchQueries: string[];
}
```

Only to get this error:

> You are assigning the primitive '"OneQuery,AnotherQuery"' to a non-primitive type 'string[]'. Use '.' binding instead?lit-plugin(no-complex-attribute-binding)(2)

As `converter` is intended to cast strings, perhaps this rule should be amended to allow ignoring properties with a `converter` or with `type: Array` or `type: Object`.

43081j:
Fixes #123 

like i mention in the issue, i might add on to this to make `no-incompatible-type-binding` also take them into account.

let me know if you think thats sensible


packages/lit-analyzer/src/rules/util/type/is-assignable-in-attribute-binding.ts
```diff
@@ -16,6 +17,15 @@ export function isAssignableInAttributeBinding(
 	const { assignment } = htmlAttr;
 	if (assignment == null) return undefined;
 
+	if (htmlAttr.kind === HtmlNodeAttrKind.ATTRIBUTE) {
```
43081j, line 20: @runem is this what you had in mind?

runem: Exactly :blush: :+1:

runem:
Btw, I see that you needed to manually run `prettier`. It seems like `lint-staged` with `husky` doesn't run for each commit on your machine. Do you have any idea why this is so?

43081j:
Good question. I just commit via git through a terminal so I would have expected husky to be fine. 

I'll have a look tomorrow and see if there's any obvious reason it isn't

43081j:
@runem is this still a valid PR? could we try get it merged if its still needed. i can catch it up, presumably its pretty far behind by now

PR #139, opened 2020-08-29
lit-analyzer: ignore type-only imports

### LINKED ISSUE #137: import type is not recognized by lit-analyzer
Hello,

we just stumbled upon the issue that lit-analyzer treats [type-only imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html#type-only-imports-and-export) as regular imports. 
Consider the following example: 
````typescript
import type {ComponentA} from 'foo/componentA' // defines <component-a>
const component: ComponentA = document.queryselector('component-a')
const template = html`<component-a></component-a>`
````
This should trigger the missing-import-rule. However, at the moment it does not. 

Thanks,
Christian

### PR DESCRIPTION
**Pull request by 43081j**
fixes #137.

this looks like it might be ok to sort the issue. let me know if you see anything wrong with it @runem 

i suppose the dependencies array contains the entry point always? and thats why i had to assert for it rather than `[]`?

### DIFF
--- a/packages/lit-analyzer/src/analyze/parse/parse-dependencies/visit-dependencies.ts
+++ b/packages/lit-analyzer/src/analyze/parse/parse-dependencies/visit-dependencies.ts
@@ -109,7 +109,7 @@ function visitDirectImports(node: Node, context: IVisitDependenciesContext): voi
 	if (node == null) return;
 
 	// Handle top level imports/exports: (import "..."), (import { ... } from "..."), (export * from "...")
-	if (context.ts.isImportDeclaration(node) || context.ts.isExportDeclaration(node)) {
+	if ((context.ts.isImportDeclaration(node) && !node.importClause?.isTypeOnly) || context.ts.isExportDeclaration(node)) {
 		if (node.moduleSpecifier != null && context.ts.isStringLiteral(node.moduleSpecifier) && context.ts.isSourceFile(node.parent)) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by 43081j** [comment]
@runem do we still need this / can we push ahead with it some time? would be nice to avoid it going stale

**Comment by justinfagnani** [comment]
@43081j I think this is a reasonable approach... a type-only import doesn't seem to fit the idea of a direct import in any sense.

File: packages/lit-analyzer/src/analyze/parse/parse-dependencies/visit-dependencies.ts [hunk 1]
```diff
@@ -109,7 +109,7 @@ function visitDirectImports(node: Node, context: IVisitDependenciesContext): voi
 	if (node == null) return;
 
 	// Handle top level imports/exports: (import "..."), (import { ... } from "..."), (export * from "...")
-	if (context.ts.isImportDeclaration(node) || context.ts.isExportDeclaration(node)) {
+	if ((context.ts.isImportDeclaration(node) && !node.importClause?.isTypeOnly) || context.ts.isExportDeclaration(node)) {
```
justinfagnani, line 112: I think you should handle type-only exports here as well.

justinfagnani (changes requested):

43081j:
@justinfagnani I've updated with a check for type only exports, if you don't mind taking another look when you get chance

justinfagnani (approved):

PR #144, opened 2020-11-01
upgrade parse5 and drop workaround types

43081j:
if one of the lurkers finally reviews my parse5 types rewrite at DefinitelyTyped/DefinitelyTyped#48402, we can drop the workaround types here and use the official ones.

draft until someone reviews the parse5 types branch, if that day ever comes.


43081j:
@runem my parse5 types got merged now so this is ready

justinfagnani:
@43081j change looks good to me, any new ideas on the build failure?

43081j:
i tried all sorts locally, im wondering if its the install magic i told you about.

it'll be installing dependencies via `npm i`, so i wonder if that means it picks up the currently published lit-analyzer as a dependency rather than the one in the repo (which has new parse5). 

just out of interest i tried bootstrapping it like anything else (knowing that wont work with vscode) and it got a bit further but still complained about an "invalid" parse5 dependency...

also weirdly enough it seems the current lit-plugin in master doesn't actually declare lit-analyzer as a dep, but directly imports it 🤔  that does't seem right, i guess it worked so far because of the npm script that copies it across

43081j:
ok i think i got it

we run the following:

* `npm run install:safe --scope lit-plugin` - which installs the dependencies of the vscode plugin from NPM, _not_ from the local monorepo
* `npm run copylink` - which then copies lit-analyzer and ts-lit-plugin's `lib` directories over the top of what was installed

This means the `node_modules` in there are from the published package still, not from the local copy. I suppose thats because if you copied the local `node_modules` it'd be full of symlinked hoisted dependencies managed by lerna (which vscode doesn't like).

Issue #165, opened 2021-04-15
[lit-analyzer] target can actually be a string


43081j:
https://github.com/runem/lit-analyzer/blob/780b3b8977d7f3dc84a544d6f6492371fea2386e/packages/lit-analyzer/src/analyze/data/extra-html-data.ts#L61

As far as im aware, these are _special_ targets the browser understands, but a target can also be any other string.

As per the spec:

> A valid browsing context name or keyword is any string that is either a valid browsing context name or that is an ASCII case-insensitive match for one of: _blank, _self, _parent, or _top.

Which means it should really be defined as `string|'_blank'|'_self'|'_parent'|'_top'` but that might as well be `string` because of the widening it causes.

43081j:
Similarly, it seems `sandbox` of the `iframe` element is too strongly typed and should be a union instead:

```html
<iframe sandbox="allow-scripts"> <!-- THIS WORKS -->

<iframe sandbox="allow-scripts allow-same-origin"> <!-- THIS DOES NOT -->
```

PR #168, opened 2021-05-01
lit-analyzer: account for string targets

43081j:
Part of #165 

A target of a link can actually be any valid string as well as the special cases.

> Otherwise, a new browsing context is being requested, and what happens depends on the user agent’s configuration and abilities — it is determined by the rules given for the first applicable option from the following list:

from here:
https://www.w3.org/TR/html52/browsers.html#the-rules-for-choosing-a-browsing-context-given-a-browsing-context-name

### DIFF
--- a/packages/lit-analyzer/src/analyze/data/extra-html-data.ts
+++ b/packages/lit-analyzer/src/analyze/data/extra-html-data.ts
@@ -58,7 +58,7 @@ const HTML_5_ATTR_TYPES: { [key: string]: string | string[] | [string[]] } = {
 	title: "string",
 	manifest: "",
 	href: "string",
-	target: ["_blank", "_parent", "_self", "_top"],
+	target: "string|\"_blank\"|\"_parent\"|\"_self\"|\"_top\"",
 	rel: "",
 	media: "",
 	hreflang: "",
--- a/packages/lit-analyzer/src/analyze/data/extra-html-data.ts
+++ b/packages/lit-analyzer/src/analyze/data/extra-html-data.ts
@@ -58,7 +58,7 @@ const HTML_5_ATTR_TYPES: { [key: string]: string | string[] | [string[]] } = {
 	title: "string",
 	manifest: "",
 	href: "string",
-	target: "string|\"_blank\"|\"_parent\"|\"_self\"|\"_top\"",
+	target: 'string|"_blank"|"_parent"|"_self"|"_top"',
 	rel: "",
 	media: "",
 	hreflang: "",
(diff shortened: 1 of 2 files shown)


packages/vscode-lit-plugin/package.json
```diff
@@ -36,7 +36,7 @@
 	"dependencies": {
 		"lit-html": "https://github.com/mjbvz/vscode-lit-html.git#1.11.1",
 		"ts-lit-plugin": "1.2.1",
-		"vscode-styled-components": "https://github.com/styled-components/vscode-styled-components#4320c223666ea835e8c63c85573c2704f550b7df",
+		"vscode-styled-components": "https://github.com/styled-components/vscode-styled-components#v1.4.0",
```
43081j, line 39: at some point vscode pushed some breaking change which means master no longer builds for me. some googling shows you need to upgrade whatever depends on vscode, which seems to be this... but we can't upgrade it to the latest version because even with `disable-scripts` it tries to run husky which fails... 🤦‍♂️

justinfagnani: I know it's very small, but could you separate the build fix into a separate PR?

43081j: i will, as my other PRs depend on it too. good to separate it

packages/lit-analyzer/src/analyze/data/extra-html-data.ts
```diff
@@ -58,7 +58,7 @@ const HTML_5_ATTR_TYPES: { [key: string]: string | string[] | [string[]] } = {
 	title: "string",
 	manifest: "",
 	href: "string",
-	target: ["_blank", "_parent", "_self", "_top"],
+	target: 'string|"_blank"|"_parent"|"_self"|"_top"',
```
justinfagnani, line 61: This just reduces to string, though I think it's nice to keep for docs. Is there some level of warning that's very low like a "did you mean?". I could see there being nice suggestions if you accidentally typed `blank` instead of `_blank` for instance.

justinfagnani (approved):

packages/lit-analyzer/src/analyze/data/extra-html-data.ts
43081j: yeah it does, i figured its good to keep the rest of the literals around for docs/autocomplete. i think normally vscode will at least show `_blank` if you start typing `blank` (as it should show it as a suggestion). but not sure how we could hint at vscode to prioritise those

PR #171, opened 2021-05-15
WIP: clean up lit-plugin scripts

43081j:
* Rename `vscode-lit-html` to what it is (it was aliased as `lit-html` before)
* Use lerna to bootstrap lit-plugin instead of a custom `install:safe` script
* Rename `compile` script of lit-plugin to `build`

Basically, this bootstraps everything normally _excluding lit-plugin_. It then bootstraps again ignoring scripts.

At this point, we have to bootstrap everything so lerna will symlink the local dependencies (ts-lit-plugin, lit-analyzer) for lit-plugin. 

We then run the copylink script which now removes the two symlinks and copies the whole package over locally (excluding node_modules).

edit:

This is still WIP i guess, the hoisting we have enabled and the lack of lock files is making this more difficult than i expected


43081j:
@justinfagnani @rictic would you be able to help me with this?

### what we did before:

1. bootstrap everything except vscode-lit-plugin **with hoisting** (all the node_modules end up in top level, no packages have lock files)
2. do a regular `npm i` in vscode-lit-plugin
3. copy the `lib/` of the local packages over the top of the ones npm install in vscode-lit-plugin

This meant that the dependencies npm had installed would be of the lit-analyzer/ts-lit-plugin in the npm registry, not of the ones locally... mismatching the sources copied across in step 3.

### what it does now

1. bootstrap everything except vscode-lit-plugin **without hoisting**, every package gets its own node_modules with its own lock files we now commit
2. bootstrap everything again with scripts disabled _including_ vscode-lit-plugin (because i couldn't find a way to only disable scripts for one local package but not the rest)
3. copy the whole ts-lit-plugin/lit-analyzer packages except `src/` across to replace the symlinks in vscode-lit-plugin

### why it doesn't work

```sh
npm list --production --parseable --depth=9999
# npm ERR! missing: lit-analyzer@^1.2.1, required by lit-plugin@1.2.1
# npm ERR! missing: ts-lit-plugin@^1.2.1 required by lit-plugin@1.2.1
```

both are definitely there inside node_modules and are not symlinks, and contain the right files.

i suspect maybe this is because the new `package-lock.json` of vscode-lit-plugin has no entry for either of these. it seems like local packages dont appear in the lock file... maybe `npm ls` expects them to be there?

ive also tried with node 16 but that resulted in another problem: for some weird reason fsevents ends up trying to install on linux and blows up the whole thing 🤦‍♂️ 

the whole reason we need two bootstraps is because of that annoying husky hook vscode-lit-html has too...

justinfagnani:
Hey @43081j I'm much more familiar (though not completely!) with the build setup now and understand these changes. I made some of these changes, are the rest still applicable?

43081j:
@justinfagnani i think there's a couple of things we still need to figure out (not necessarily in this pr):

* rename the `compile` script to `build` for the vscode plugin
* add lit-analyzer as a dependency of the vscode plugin ([see here](https://github.com/runem/lit-analyzer/blob/1e1267dc1da0d58c90cb25b8fa4ae48f9f6fb377/packages/vscode-lit-plugin/src/extension.ts#L1))
* disable hoisting

The big one is disabling hoisting, it causes a lot of weirdness especially combined with us manually overwriting dependencies inside the vscode plugin. We could really do with figuring that out once and for all, a way to have the vscode plugin depend on the LOCAL packages (lit-analyzer, ts-lit-plugin) without manually copying stuff around. 

the copy script means we often run into problems with `vsce publish` which uses npm to check the dependency tree's validity. it often tells us the tree is invalid because the copied across dependency isn't the one it was expecting. this is worsened when you add package locks it seems, as it validates against the lock file (which the copied package isn't valid against).

all this copying stuff exists because vsce requires all the dependencies to exist in the package's node_modules, _no symlinks allowed_. so we can't just symlink our local packages.

this also means we have a deadlock anytime we want to update a package. lets say you update lit-analyzer and want to run tests and package the extension, you won't often be able to because it'll want the published lit-analyzer rather than your local one.

rictic:
this is now fixed at head!

Issue #141, opened 2020-09-25
lit-html 2.0 support


justinfagnani:
We're working on lit-html 2.0 right now in a branch here: https://github.com/Polymer/lit-html/tree/lit-next
It's also published to npm with the `@next-major` tag: https://www.npmjs.com/package/lit-html/v/2.0.0-pre.3

There aren't a great many changes that would effect analysis of templates. I think the biggest change is in directives, which are class-based and might be generically analyzable now based on their render method.

Some very small things:
- Symbols can no longer be passed to attribute and text bindings, they have to be converted to strings first.
- Arrays passed to attributes are not joined with a space. Authors need to convert them to strings.
- Dynamic template tags are not allowed. It was extremely rare anyway.
- TemplateResult is no longer a class. It's an object differentiated with a special property.
- DirectiveResult is now a thing and similar to TemplateResult. It includes a reference to the Directive class

Bigger:
- We're probably going to allow "element" or "spread" bindings, but just with directives, like
```ts
html`<div ${ref()}></div>` 
```
   (TBD on the ref part, but that would be a directive)
- We're probably going to add static "bindings", which are interpolated into the template before bindings. These can't be updated, so should usually be const module variables.
- Directives are classes. Their signature is derived from a `render()` methods (name might change) that returns another lit-html type. ie, a directive for a text binding can now return a template result.

Opening this issue to let you know and track anything that might need to be done.

Thanks!

Artur-:
`<vaadin-text-field ${field(this.binder.model.name)}></vaadin-text-field>` now works with lit 2.0.0-pre.1 but from this plugin I get 
```
Unknown attribute '________________________________'.
```

would be cool to get lit 2 support

Christian24:
Maybe a new rule for for lit 2.0 would be nice to make people aware of legacy imports like 

`import {html} from 'lit-html'` 
or 
`import {LitElement} from 'lit-element'`

and change it to 
`import {LitElement, html} from 'lit`.

e111077:
Directives should be partially addressed in #172

PR #172, opened 2021-05-20
feat(directives): Recognize lit 2 directives

e111077:
Lit 2 the signature of the directive function is not `directive<T extends typeof DirectiveClass>(T) => DirectiveResult<T>`

This PR adds recognition of lit 2 directives but currently does not implement value inferencing, though it should be possible since it's just the return type of the `render` function of `T`

Related #141


e111077:
review request: @justinfagnani

e111077:
ping @rictic @justinfagnani

justinfagnani:
I did this in #189 but with separate Lit 1 and Lit 2 detection functions so we can require Lit 2 directives for element expressions.

PR #174, opened 2021-06-07
[lit-analyzer] remove circular analyzer dependency

### LINKED ISSUE #173: Remove the circular dependency in lit-analyzer
Inside lit-analyzer, you will see this:

https://github.com/runem/lit-analyzer/blob/9aceb3d340b8a569938318c1f94bd997eaa89957/packages/lit-analyzer/src/analyze/util/rule-fix-util.ts#L2

This means lit-analyzer depends on ts-lit-plugin, which depends on lit-analyzer...

Currently, it works because rollup figures it out. Really, though, it shouldn't be a thing, `ts-module` should likely live in the analyzer.

43081j:
Fixes #173 

Seems it may have just been a typo or something. This will allow us to drop rollup eventually


justinfagnani:
Just hit this myself when removing Rollup

justinfagnani (approved):

Issue #45, opened 2019-08-16
Rollup necessity? 


43081j:
out of interest, is there a reason you use rollup?

why not just build directly using `tsc` with CJS/ES2015+ as a target? (which would work fine in node)?

runem:
I have also been thinking about removing rollup and just going with `tsc`, especially because there is a noticeable difference in speed between them.

**Here are my reasoning for introducing rollup:**
1. I planned on generating 2 separate bundles: "cli" and "analyzer".
2. The rollup setup merges Typescript declarations (for convenience)
3. I bake in the version of the analyzer when it's built see [constants.ts](https://github.com/runem/lit-analyzer/blob/master/packages/lit-analyzer/src/analyze/constants.ts#L19). This constant is used for debugging and for printing the version of the CLI.

I'm open for just using `tsc`, but we would have to either remove `VERSION´ or find another way to get the version of the package.

justinfagnani:
Rollup will probably be a win for startup in Node, even if it's not required. Of course, an app can do that too.

From my side of the world, I'd love to see a pure modules distribution for use in something like: https://github.com/PolymerLabs/code-sample-editor

justinfagnani:
In trying to [upgrade to modules](https://github.com/runem/lit-analyzer/issues/184) I'm running into a lot of problems related to ts-node and Ava and how test files import from source directly.

I think a lot of this should be simplified by just using `tsc` and having all tooling downstream from that only consume .js files.

Bundling can be done at the app entry points like the VS Code extension and the CLI, but the analyzer _library_ probably shouldn't be bundled. The TypeScript declarations shouldn't need to be merged either.

I'll put together a PR with this and migration to modules. It'll probably be large, but I think I have to try out all the changes together to see if they work.

Issue #184, opened 2021-07-09
[lit-analyzer] Upgrade to modules


justinfagnani:
Since lit-html is published as modules, using utilities from it, such as `getTemplateHTML` will be easier it lit-analyzer uses modules too.

Hopefully there's no issue with using modules inside a VS Code extensions.

PR #185, opened 2021-07-09
[lit-analyzer] Remove Rollup and build with plain tsc

justinfagnani:
This simplifies the build pipeline and prepares the packages to use native modules (https://github.com/runem/lit-analyzer/issues/184).

I _strongly_ recommend reviewing this PR commit-by-commit. I crafted and ordered each commit to be easies to review. This means that the branch doesn't build correctly until the 4th commit in.

* Moves all .ts files, including tests, into /src/
* Compiles with `tsc` rather than rollup
* Adds file extensions to imports and enforces with eslint
* Removes ts-node: Ava is testing the JS output of tsc
* The version string in constants.ts is now updated manually, but checked with a `check-version` script.

Fixes https://github.com/runem/lit-analyzer/issues/45


justinfagnani:
cc @43081j

justinfagnani:
I don't understand the CI failure. It says "Cannot find module 'typescript'" even though it's a dev dep in all package.json's including the root which should be sufficient for eslint. I added "typescript" to the nohoist config in lerna.json, but it should still be installed in the top-level node_modules/ folder.

43081j:
it'll be the v2 package lock i think. i had the same problem in another repo yesterday in fact and changing my ci to use node 16 solved the problem.

the solution shouldn't be to change CI to use node 16, though, as this project should build on 14 too. i couldn't figure out another solution though.

it seems node <16 just really doesn't handle new lockfiles well. often gets the node_modules tree in a weird state where some dependencies are missing completely.

43081j (approved):

aomarks (approved):
Everything looks good to me! +1 to shifting to direct tsc output. Accolades for splitting into logical commits.

43081j:
CI failure may be unrelated but this is one of the missing dependencies i was talking about @justinfagnani 

vscode-lit-plugin depends on lit-analyzer in its sources, but doesn't list it as a dependency in package.json. might cause some funky stuff

justinfagnani:
Thanks @43081j! Yeah, I unblocked the build failure with installing npm@7, then a lint failure, and now this dep issue. Hopefully it's the last one.

43081j:
@justinfagnani if you can figure this one out, you're a wizard 😂 

https://github.com/runem/lit-analyzer/pull/171#issuecomment-841681948

see i ran into the exact same error and lost my mind trying to find out why

justinfagnani:
It looks like maybe npm@7 doesn't handle Git dependencies in `npm ls`?

justinfagnani:
No, that's not it. I made a small repro package and `npm ls` worked fine with npm@7.19.0.

43081j:
> npm verb shrinkwrap failed to load node_modules/.package-lock.json out of date, updated: node_modules/lit-analyzer

maybe this is relevant?

i  had no idea there was such a thing as a `node_modules/.package-lock.json`

43081j:
because we will have copied over the package.json from our local repo, which replaced the one npm generated originally (presumably had some extra information in it?)

justinfagnani:
I tried deleting `node_modules/.package-lock.json` and the result was the same

justinfagnani:
The annoying thing is that npm is not saying why the package is invalid. The package names and versions match (this is why `vscode-lit-html` is installed as `lit-html` - it's package.json name is `lit-html`). The thing I'm looking for now is some other reference to `lit-html` that's expecting the actual library.

43081j:
@justinfagnani 

```ts
      // if it's a git type, verify that they're the same repo
      //
      // if it specifies a definite commit, then it must have the
      // same commit to be considered the same repo
      //
      // if it has a #semver:<range> specifier, verify that the
      // version in the package is in the semver range
```

_verify that they're the same repo_ i think is why it fails

so i suspect its because the package.json we copied across has no `resolved` containing a github url. ofc don't want to hack that in though

justinfagnani:
Ah, I should have looked here. I just debugged it to the same place. I really wish npm had included all these reasons for an edge to be invalid in the error. The problem here is that `resRepo.fetchSpec` is `"latest"`, but `requested.fetchSpec` is `"https://github.com/mjbvz/vscode-lit-html.git"`. 

Does this relate to the package.json's that are copied? Those are for `lit-analyzer` and `ts-lit-plugin`, are there more?

justinfagnani:
Argh. Using GitHub dependencies fixed `npm run package` locally for me - it had been failing the same way as on CI before.

justinfagnani:
Success!!!

43081j:
just to double check, doesn't that change mean we will be re-installing its dependencies, i.e. re-overwriting the ones we copied using the copylink script?

as in:

* install dependencies
* overwrite ts-lit-plugin and lit-analyzer with local copies
* install dependencies <- here have we now just undone the overwrite from the previous step?

justinfagnani:
@43081j these steps do seem to install the local lit-analyzer. I'm able to test the plugin with changes this way.

PR #188, opened 2021-08-12
Add support for Lit 2 element expressions

justinfagnani:
This adds support for expressions in "element position":

```ts
html`<input ${ref(inputRef)}>`
```

These expressions are written in attribute name position, so a number of assumptions in the code base change. We need to generate unique substitutions so that multiple element expressions per element are supported. We also need to use the attribute name location object in several places because the attribute assignment location object doesn't exist. And we need a new assignment type to differentiate between element expressions and unbound boolean attributes.

This PR doesn't not update `isLitDirective()` to detect Lit 2 directive results. That's a follow-on task and when that's done this should be updated to _only_ only Lit 2 directives and not lit-html 1.x directives.

cc @43081j and @aomarks I can't add you as reviewers, but your feedback would be very appreciated!

### DIFF
--- a/packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/parse-html-node/parse-html-attr-assignment.ts
+++ b/packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/parse-html-node/parse-html-attr-assignment.ts
@@ -28,7 +28,7 @@ export function parseHtmlAttrAssignment(
 			// in HTML. The parts will be in the range of the attribute name instead.
 			const values = context.getPartsAtOffsetRange(htmlAttr.location);
 			return {
-				kind: HtmlNodeAttrAssignmentKind.ELEMENT,
+				kind: HtmlNodeAttrAssignmentKind.ELEMENT_EXPRESSION,
 				htmlAttr,
 				location: htmlAttr.location,
 				expression: values[0] as Expression
--- a/packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/parse-html-node/parse-html-attr-assignment.ts
+++ b/packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/parse-html-node/parse-html-attr-assignment.ts
@@ -1,3 +1,4 @@
+import { Expression } from "typescript";
 import { HtmlNodeAttrAssignment, HtmlNodeAttrAssignmentKind } from "../../../../../types/html-node/html-node-attr-assignment-types.js";
 import { HtmlNodeAttr } from "../../../../../types/html-node/html-node-attr-types.js";
 import { Range } from "../../../../../types/range.js";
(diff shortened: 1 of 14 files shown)


packages/lit-analyzer/src/lib/rules/util/type/is-assignable-in-element-binding.ts
```diff
@@ -0,0 +1,27 @@
...
+ * value for element expressions.
+ *
+ * Note: This currently checks against the lit-html 1.x interface for a
+ * directive. When lit-analyzer can detect Lit 2 directives it will check that
+ * the directive is specifically a Lit 2 directive.
+ */
+export function isAssignableInElementBinding(htmlAttr: HtmlNodeAttr, type: SimpleType, context: RuleModuleContext): boolean | undefined {
+	// TODO (justinfagnani): is there a better way to determine if the
+	// type *contains* any, rather than *is* any?
+	if (!isLitDirective(type) && !(type.kind === "ANY")) {
```
aomarks, line 18: nit: `type.kind !== "ANY"`

packages/lit-analyzer/src/lib/analyze/types/html-node/html-node-attr-assignment-types.ts
```diff
@@ -6,7 +6,8 @@ export enum HtmlNodeAttrAssignmentKind {
 	BOOLEAN = "BOOLEAN",
 	EXPRESSION = "EXPRESSION",
 	STRING = "STRING",
-	MIXED = "MIXED"
+	MIXED = "MIXED",
+	ELEMENT = "ELEMENT"
```
aomarks, line 10: would `ELEMENT_EXPRESSION` be more clear?

aomarks (approved):
Nice!

Just a few minor suggestions/questions.

packages/lit-analyzer/src/lib/analyze/parse/document/virtual-document/virtual-ast-document.ts
```diff
@@ -129,8 +130,15 @@ export class VirtualAstDocument implements VirtualDocument {
 		}
 	}
 
-	protected substituteExpression(length: number, expression: Expression, prev: string, next: string | undefined): string {
-		return "_".repeat(length);
+	protected substituteExpression(length: number, expression: Expression, prev: string, next: string | undefined, index: number): string {
```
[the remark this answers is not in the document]
justinfagnani: That hasn't been done elsewhere in the project and the `noUnusedParameters` isn't set, so I'd prefer to leave this as-is and come back to that on a compiler options / linting update for the whole project.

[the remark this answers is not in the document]
aomarks: Oh ok. You did add an `_index` parameter in `virtual-css-document.ts` though.

packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/parse-html-node/parse-html-attr-assignment.ts
```diff
@@ -20,6 +21,19 @@ export function parseHtmlAttrAssignment(
 	const location = getAssignmentLocation(p5Node, p5Attr, htmlAttr, context);
 
 	if (location == null) {
+		// A null assignment location might be an element expression, which only
+		// has an attribute name and no attribute "assignment".
+		if (htmlAttr.name.match(/_+\d+_/)) {
```
aomarks, line 26: I guess you'll also need to update this to `/_+[0-9a-z]+_/` to account for the base 10 -> 36 change.

justinfagnani: oops. I need a test to hit that...

justinfagnani: done

aomarks (approved):

justinfagnani:
@aomarks I had to modify two more rules to ignore element expressions. I've tested this in VS Code and lit-html 1.x directives now have no errors in element expressions. Fixing directive detection is the next priority. PTAL at the last commit?

aomarks (approved):

aomarks:
> @aomarks I had to modify two more rules to ignore element expressions. I've tested this in VS Code and lit-html 1.x directives now have no errors in element expressions. Fixing directive detection is the next priority. PTAL at the last commit?

LGTM

PR #189, opened 2021-08-13
Recognize Lit 2 directives

justinfagnani:
This adds `isLit1Directive()` and `isLit2Directive()` and makes `isLitDirective()` return true if either does. Most call sites are left alone so they detect either Lit 1 or Lit 2 directives. The `isAssignableInElementBinding()` helper checks for a Lit 2 directive specifically.

### DIFF
--- a/packages/lit-analyzer/package.json
+++ b/packages/lit-analyzer/package.json
@@ -46,7 +46,7 @@
 		"didyoumean2": "4.1.0",
 		"fast-glob": "^2.2.6",
 		"parse5": "5.1.0",
-		"ts-simple-type": "2.0.0-next.0",
+		"ts-simple-type": "~1.0.5",
 		"vscode-css-languageservice": "4.3.0",
 		"vscode-html-languageservice": "3.1.0",
 		"web-component-analyzer": "~1.1.1"
--- a/packages/lit-analyzer/package.json
+++ b/packages/lit-analyzer/package.json
@@ -46,7 +46,7 @@
 		"didyoumean2": "4.1.0",
 		"fast-glob": "^2.2.6",
 		"parse5": "5.1.0",
-		"ts-simple-type": "~1.0.5",
+		"ts-simple-type": "2.0.0-next.0",
 		"vscode-css-languageservice": "4.3.0",
 		"vscode-html-languageservice": "3.1.0",
 		"web-component-analyzer": "~1.1.1"
--- a/packages/lit-analyzer/src/lib/rules/util/directive/is-lit-directive.ts
+++ b/packages/lit-analyzer/src/lib/rules/util/directive/is-lit-directive.ts
@@ -10,11 +10,16 @@ const partTypeNames: ReadonlySet<string | undefined> = new Set([
 ]);
 
 /**
- * Checks whether a type is a lit directive.
- * It will return true if the type is a function that takes a Part type and returns a void.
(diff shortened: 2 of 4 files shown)


justinfagnani:
cc @aomarks @43081j

packages/lit-analyzer/package.json
```diff
@@ -46,7 +46,7 @@
 		"didyoumean2": "4.1.0",
 		"fast-glob": "^2.2.6",
 		"parse5": "5.1.0",
-		"ts-simple-type": "~1.0.5",
+		"ts-simple-type": "2.0.0-next.0",
```
aomarks, line 49: I assume this is intentionally an `=` equivalent instead of `^` because pre-releases are unstable?

packages/lit-analyzer/src/lib/rules/util/directive/is-lit-directive.ts
```diff
@@ -10,11 +10,16 @@ const partTypeNames: ReadonlySet<string | undefined> = new Set([
...
+	return isLit1Directive(type) || isLit2Directive(type);
+}
+
+/**
+ * Checks whether a type is a lit-html 1.x directive.
+ */
+export function isLit1Directive(type: SimpleType): boolean {
 	switch (type.kind) {
 		case "ALIAS":
 			return type.name === "DirectiveFn" || isLitDirective(type.target);
```
aomarks, line 25: Should this be `|| isLit1Directive(type.target)` ?

packages/lit-analyzer/src/lib/rules/util/type/is-assignable-in-element-binding.ts
```diff
@@ -15,11 +15,18 @@ import { isLitDirective } from "../directive/is-lit-directive.js";
 export function isAssignableInElementBinding(htmlAttr: HtmlNodeAttr, type: SimpleType, context: RuleModuleContext): boolean | undefined {
```
aomarks, line 15: Update the comment above about lit 1 vs 2

aomarks (approved):

packages/lit-analyzer/package.json
aomarks: Also why did we need the bump?

43081j (approved):

packages/lit-analyzer/src/lib/rules/util/directive/is-lit-directive.ts
justinfagnani: oh, yes

packages/lit-analyzer/package.json
justinfagnani: I was trying to see if a newer ts-simple-type would fix the error I was seeing. This should have been bumped anyway.

justinfagnani: I'm going to back this out and see if it fixes the build. For some reason `npm run bootstrap` isn't regenerating package locks and I think that's causing `npm ls` to fail and I don't want to deal with that now.

justinfagnani:
I had to add a little more logic to `isLit2Directive` to make it work with live Lit 2 code. The tests aren't exact since I had to remove the generic from `DirectiveResult` to avoid a ts-simple-type crasher.

aomarks (approved):

PR #190, opened 2021-08-17
Rename vscode-lit-plugin compile script to "build"

justinfagnani:
This way `npm run build` from the root builds the plugin too.


rictic:
This is fixed at head

PR #199, opened 2021-09-15
Update WCA, typescript and internal dependencies

justinfagnani:
Also bumps version number to prep for some `@next` releases.


rictic (approved):

rictic:
Some of this was superceded by #206 but it looks like there's still other changes here that are worth landing

melink14:
Since this PR has the goal of updating dependencies I wonder if it should also update fast-glob as per https://github.com/runem/lit-analyzer/issues/226 or if that should be in it's own PR...?

rictic:
Updating fast-glob in #244

PR #210, opened 2021-11-03
no-incompatible-property-type: ignore @property(variable)

### LINKED ISSUE #209: @property(someVariable) unwanted warning on
Given this code:

```js
  @property(dateProperty) date: Temporal.PlainDate = Temporal.Now.plainDateISO();
```

I'm getting a warning in VSCode.

> The built in converter doesn't handle the property type 'PlainDate'. Please add '{attribute: false}' on @property decorator for 'date'lit-plugin(no-incompatible-property-type)(2303)

For what it's worth, here's the definition of dateProperty:

```js
const dateProperty = {
  reflect: true,
  converter: {
    fromAttribute(string: string): Temporal.PlainDate {
      return Temporal.PlainDate.from(string);
    },
    toAttribute(plainDate: Temporal.PlainDate): string {
      return plainDate.toString();
    },
  },
};
```

I'm not expecting the plugin to find the definition of `dateProperty` and see that it has a converter. I'm expecting it to say, "well, there's no way for me to know for sure" and not emit an error.

pmcelhaney:
Meager attempt to fix #209

### DIFF
--- a/packages/lit-analyzer/src/lib/rules/no-incompatible-property-type.ts
+++ b/packages/lit-analyzer/src/lib/rules/no-incompatible-property-type.ts
@@ -17,6 +17,11 @@ const rule: RuleModule = {
 
 		if ((member.meta.node?.type ?? member.node)?.getSourceFile() !== context.file) return;
 
+		// In case of @property(someVariable) we can't check
+		if (member.meta?.node?.decorator?.arguments?.[0]?.constructor?.name === "IdentifierObject") {
+			return;
+		}
+
 		// Grab the type and fallback to "any"
 		const type = member.type?.() || { kind: "ANY" };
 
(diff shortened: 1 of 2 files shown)


packages/lit-analyzer/src/lib/rules/no-incompatible-property-type.ts
```diff
@@ -127,7 +132,7 @@ function validateLitPropertyConfig(
 	}
 
 	// Don't continue if we don't know the property type (eg if we are in a js file)
-	// Don't continue if this property has a custom converter (because then we don't know how the value will be converted)
+	// Don't continue if this proper
```
pmcelhaney, line 135:
```suggestion
	// Don't continue if this property has a custom converter (because then we don't know how the value will be converted)
```

43081j:
This will be fixed by  runem/web-component-analyzer#233, so we should close this PR.

contribution is appreciated though

PR #231, opened 2022-02-03
Log to the TS server log too.

rictic:


rictic:
We found it easier to debug the plugin when its logs were mixed in with the TS server logs too

justinfagnani (approved):

PR #232, opened 2022-02-03
Add an integration test of completions.

rictic:


packages/vscode-lit-plugin/src/test/fixtures/completions.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+customElements.define("complete-me", CompleteMe);
+declare global {
+	interface HTMLElementTagNameMap {
+		"complete-me": CompleteMe;
+	}
+}
+
+html`
+	<complete-me
```
justinfagnani, line 22: Why's this formatted this way? Is the completion done at the end of the first line? Maybe add a comment to day that.

justinfagnani (approved):

packages/vscode-lit-plugin/src/test/fixtures/completions.ts
rictic: Added comment

PR #240, opened 2022-03-31
Use ...args to ensure we pass all args through

rictic:
The language service we're overriding may be from a different / future version of typescript, and so may take additional parameters.


AndrewJakubowicz (approved):
Awesome!

PR #241, opened 2022-04-06
Use wireit to run scripts

rictic:
This lets scripts specify their dependencies, avoid re-doing work unnecessarily, and has a robust watch mode for free!

@aomarks

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,29 +27,67 @@
 	],
 	"scripts": {
 		"postinstall": "npm run bootstrap",
-		"prettier:check": "prettier --check \"packages/*/src/**/*.ts\"",
+		"prettier:check": "wireit",
 		"prettier:write": "prettier --write \"packages/*/src/**/*.ts\"",
 		"postpublish": "lerna run install:safe --scope lit-plugin && echo \"🎉 Published successfully! You can now publish the vscode extension seperately by running 'cd packages/vscode-lit-plugin && npm run publish'\"",
 		"publish": "lerna run prepare && lerna publish --exact && npm run postpublish",
 		"publish:next": "lerna run prepare && echo \"When asked, please choose 'Custom Prerelease' and use default value\" && lerna publish --dist-tag next --preid next --exact && npm run postpublish",
 		"bootstrap": "lerna clean --yes && lerna bootstrap --hoist --ignore lit-plugin && lerna run install:safe --scope lit-plugin && npm run copylink",
-		"lint": "eslint .",
-		"build": "lerna run build --stream",
+		"lint": "wireit",
+		"build": "wireit",
 		"watch": "lerna run watch --parallel --stream",
-		"test": "lerna run test --parallel --stream",
-		"test:all": "lerna run test:all --parallel --stream",
+		"test": "wireit",
 		"test:watch": "lerna run test:watch --parallel --stream",
 		"readme": "lerna run readme && readme generate -i readme.blueprint.md -c readme.config.json",
 		"dev": "cd dev && TSS_DEBUG=5999 code . --disable-extension runem.lit-plugin",
 		"dev:logs": "touch dev/lit-plugin.log && tail -f dev/lit-plugin.log",
-		"copylink": "node scripts.js copylink",
-		"copylink:watch": "nodemon --exec \"npm run copylink\""
+		"copylink": "wireit",
+		"copylink:watch": "nodemon --exec \"npm run copylink\"",
+		"just-package": "wireit"
+	},
+	"wireit": {
+		"build": {
+			"dependencies": [
+				"./packages/lit-analyzer:build",
+				"./packages/ts-lit-plugin:build",
+				"./packages/vscode-lit-plugin:build"
+			]
+		},
+		"test": {
+			"dependencies": [
+				"./packages/lit-analyzer:test",
+				"./packages/vscode-lit-plugin:test",
+				"lint",
+				"prettier:check"
+			]
+		},
+		"copylink": {
+			"dependencies": [
+				"./packages/lit-analyzer:build",
+				"./packages/ts-lit-plugin:build"
+			],
+			"files": [
+				"packages/lit-analyzer",
+				"!packages/lit-analyzer/node_modules",
+				"!packages/lit-analyzer/.wireit",
+				"packages/ts-lit-plugin",
+				"!packages/ts-lit-plugin/node_modules",
+				"!packages/ts-lit-plugin/.wireit"
+			],
+			"command": "node scripts.js copylink"
+		},
+		"lint": {
+			"command": "eslint ."
+		},
+		"prettier:check": {
+			"command": "prettier --check \"packages/*/src/**/*.ts\""
+		}
 	},
 	"devDependencies": {
 		"@appnest/readme": "^1.2.7",
-		"@vscode/test-electron": "^2.1.1",
 		"@typescript-eslint/eslint-plugin": "^5.0.0",
 		"@typescript-eslint/parser": "^5.0.0",
+		"@vscode/test-electron": "^2.1.1",
 		"eslint": "^8.0.0",
 		"eslint-config-prettier": "^8.3.0",
 		"eslint-plugin-import": "^2.25.2",
@@ -59,7 +97,8 @@
 		"lint-staged": "^10.2.10",
 		"nodemon": "^2.0.4",
 		"prettier": "^2.4.1",
-		"typescript": "~4.4.3"
+		"typescript": "~4.4.3",
+		"wireit": "^0.0.0"
 	},
 	"husky": {
 		"hooks": {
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -27,29 +27,67 @@
...
-		"test:all": "lerna run test:all --parallel --stream",
+		"test": "wireit",
 		"test:watch": "lerna run test:watch --parallel --stream",
 		"readme": "lerna run readme && readme generate -i readme.blueprint.md -c readme.config.json",
 		"dev": "cd dev && TSS_DEBUG=5999 code . --disable-extension runem.lit-plugin",
 		"dev:logs": "touch dev/lit-plugin.log && tail -f dev/lit-plugin.log",
-		"copylink": "node scripts.js copylink",
-		"copylink:watch": "nodemon --exec \"npm run copylink\""
+		"copylink": "wireit",
+		"copylink:watch": "nodemon --exec \"npm run copylink\"",
```
aomarks, line 45: Could this just now be replaced with `npm run copylink watch`? Then could drop the `nodemon` dependency too.

packages/vscode-lit-plugin/package.json
```diff
@@ -25,16 +25,76 @@
...
 		"test-packaged": "vsce package -o ./out/packaged.vsix && rm -rf ../../../packaged-extension/ && mkdir ../../../packaged-extension/ && unzip -qq ./out/packaged.vsix -d ../../../packaged-extension/ && node ./out/test/scripts/test-runner.js ../../../packaged-extension/extension/"
 	},
+	"wireit": {
+		"build": {
+			"dependencies": [
+				"../ts-lit-plugin:build",
+				"../lit-analyzer:build",
+				"../../:copylink"
+			],
+			"command": "tsc --build",
```
aomarks, line 48: You'll want `"clean": false` here, otherwise the output will get deleted on every run. If the `.tsbuildinfo` is in `output`, that would work, but there would be no advantage to `--build`. If it's not in `output`, I think it will actually be broken, because `tsc` thinks previous up-to-date output exists that actually doesn't.

So you want `"clean": false` and to include the `.tsbuildinfo` file in `output`. When https://github.com/google/wireit/issues/70 lands, you can then switch to `"clean": "on-delete"`, which will give you the best result because you'll get incremental build on changes + new files, and a clean build on deleted files (which is better than the default behavior of `tsc --build`, which just leaves compiled output for deleted source files lying around).

packages/lit-analyzer/package.json
```diff
@@ -26,14 +26,53 @@
...
+		"test": "wireit",
+		"test:ava": "wireit",
+		"test:all": "npm test",
 		"test:watch": "ava --watch --color",
-		"check-version": "node scripts/check-version.js"
+		"check-version": "wireit"
+	},
+	"wireit": {
+		"build": {
+			"command": "tsc --build",
```
aomarks, line 39: See above comment. Add `"clean": false` and add `.tsbuildinfo` to `output`.

packages/ts-lit-plugin/package.json
```diff
@@ -24,9 +24,30 @@
...
+		"build": "wireit",
+		"test": "wireit",
 		"readme": "readme generate -i readme.blueprint.md -c readme.config.json"
 	},
+	"wireit": {
+		"build": {
+			"dependencies": [
+				"../lit-analyzer:build"
+			],
+			"command": "tsc --build",
```
aomarks, line 36: See above comment. Add `"clean": false` and add `.tsbuildinfo` to `output`.

package.json
```diff
@@ -27,29 +27,67 @@
...
+				"prettier:check"
+			]
+		},
+		"copylink": {
+			"dependencies": [
+				"./packages/lit-analyzer:build",
+				"./packages/ts-lit-plugin:build"
+			],
+			"files": [
+				"packages/lit-analyzer",
```
aomarks, line 70: This doesn't work at the moment -- for input files, including a directory does not implicitly include the files inside it. You'll need to write e.g. `packages/lit-analyze/**/*.{ts,js,}` for now (planning a fix for this soon).

aomarks (changes requested):

package.json
rictic: yes, and probably many of the other watch scripts can be removed too, just need to evaluate them more

ava's watch mode is potentially useful though, the ava tests take a little while

packages/lit-analyzer/package.json
```diff
@@ -26,14 +26,55 @@
...
+			"output": [
+				"lib",
+				"test",
+				"scripts",
+				"index.js",
+				"index.d.ts",
+				"index.d.ts.map",
+				"tsconfig.tsbuildinfo"
+			],
+			"clean": false
```
aomarks, line 53: Recommend `"clean": "if-file-deleted"` (needs `wireit@0.1.1`).

See https://github.com/google/wireit#typescript for explanation

rictic: Done, that's nicer

packages/lit-analyzer/package.json
```diff
@@ -26,14 +26,55 @@
...
+		"test": "wireit",
+		"test:ava": "wireit",
+		"test:all": "npm test",
 		"test:watch": "ava --watch --color",
-		"check-version": "node scripts/check-version.js"
+		"check-version": "wireit"
+	},
+	"wireit": {
+		"build": {
+			"command": "tsc --build",
```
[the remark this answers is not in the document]
rictic: Done

package.json
```diff
@@ -27,29 +27,74 @@
...
+			]
+		},
+		"copylink": {
+			"dependencies": [
+				"./packages/lit-analyzer:build",
+				"./packages/ts-lit-plugin:build"
+			],
+			"command": "node scripts.js copylink"
+		},
+		"eslint": {
```
aomarks, line 86: This is fine, but just FYI these could also be regular old npm scripts, since they dont have any dependencies, files, output. Wireit scripts can depend on non-Wireit scripts.

packages/lit-analyzer/package.json
```diff
@@ -25,15 +25,55 @@
...
+				"tsconfig.json"
+			],
+			"output": [
+				"lib",
+				"test",
+				"scripts",
+				"index.js",
+				"index.d.ts",
+				"index.d.ts.map",
+				"./.tsbuildinfo"
```
aomarks, line 50:
```suggestion
				".tsbuildinfo"
```

packages/lit-analyzer/package.json
```diff
@@ -25,15 +25,55 @@
...
+			"command": "ava --color",
+			"output": [],
+			"files": []
+		},
+		"check-version": {
+			"dependencies": [
+				"build"
+			],
+			"files": [
+				"package.json"
```
aomarks, line 73: Assuming this is not a build artifact:

```suggestion
				"package.json",
				"scripts/check-version.js"
```

packages/vscode-lit-plugin/package.json
```diff
@@ -25,31 +25,120 @@
...
+		"test:normal": {
+			"dependencies": [
+				"build"
+			],
+			"files": [
+				"package.json",
+				"schemas/**/*"
+			],
+			"output": [],
+			"command": "node ./out/test/scripts/test-runner.js"
```
aomarks, line 75: Looks like this is a build artifact, so it doesn't need to be in `files`. (Just confirming?)

Same for the other scripts below.

packages/vscode-lit-plugin/package.json
```diff
@@ -25,31 +25,120 @@
...
+		},
+		"create-isolated-extension-dir": {
+			"dependencies": [
+				"package"
+			],
+			"files": [
+				"./out/packaged.vsix"
+			],
+			"output": [
+				"../../packaged-extension/extension/**/*"
```
aomarks, line 118: Currently it's forbidden to have output outside of the package directory. It will crash when you try to do a clean build (just as an intentional check, because it seemed potentially indicative of an error) or cache (because we'll need a different directory layout to support this).

If you don't clean or cache here, it's OK though.

packages/vscode-lit-plugin/src/test/scripts/test-runner.ts
```diff
@@ -23,6 +23,10 @@ async function main() {
 		const fixturesDir = path.join(__dirname, "..", "..", "..", "src", "test", "fixtures");
 		// Download VS Code, unzip it and run the integration test
 		await runTests({ extensionDevelopmentPath, extensionTestsPath, launchArgs: [fixturesDir] });
+
+		setTimeout(function () {
```
aomarks, line 27: Is this a leftover?

Seems like it could cause a hanging test to exit with a success code and report a false success?

aomarks (changes requested):

43081j:
TIL wireit is a thing. is there a readme anywhere for it? would be good to understand how it works more

also nice to see @aomarks finally got pulled into this 😂

aomarks (approved):

PR #242, opened 2022-04-10
Bundle ts-lit-plugin and vscode-lit-plugin

### LINKED ISSUE #243: Bundle the vscode plugin
Currently we publish a ton of node modules and many files. As a result the plugin package is pretty large, 25 to 30MB, and could be faster to initialize

If we bundle the code we can have a smaller extension that's faster to init

rictic:
This greatly simplifies the process of building the vscode plugin. We construct a directory at `./packages/vscode-lit-plugin/built` which contains only those files that should be part of the plugin. The only JS that's included is:
  - A ts-lit-plugin bundle, packaged in such a way that the typescript compiler will find it
  - A bundle for the main vscode extension process
  - The TypeScript compiler, separated out for the vscode extension to load

This brings down our total vsix size from ~25MB to ~15MB, and if we trim the typescript package a bit more, I bet we can get that down to ~8MB.

This also sets the stage to try out npm workspaces instead of lerna

Fixes #243


43081j:
are you sure? when i run it i get:

* `bundle-esbuild.js` - 1.1M
* `bundle-rollup.js` - 5.2M

🤔

rictic:
I was not able to repro the 15MB number! After checking this branch out again the esbuild output has remained significantly smaller than the rollup output

package.json
```diff
@@ -27,23 +27,22 @@
 	],
 	"scripts": {
 		"postinstall": "npm run bootstrap",
-		"prettier:check": "prettier --check \"packages/*/src/**/*.ts\"",
 		"prettier:write": "prettier --write \"packages/*/src/**/*.ts\"",
-		"postpublish": "lerna run install:safe --scope lit-plugin && echo \"🎉 Published successfully! You can now publish the vscode extension seperately by running 'cd packages/vscode-lit-plugin && npm run publish'\"",
+		"postpublish": "echo \"🎉 Published successfully! You can now publish the vscode extension seperately by running 'cd packages/vscode-lit-plugin && npm run publish'\"",
 		"publish": "lerna run prepare && lerna publish --exact && npm run postpublish",
 		"publish:next": "lerna run prepare && echo \"When asked, please choose 'Custom Prerelease' and use default value\" && lerna publish --dist-tag next --preid next --exact && npm run postpublish",
-		"bootstrap": "lerna clean --yes && lerna bootstrap && lerna run install:safe --scope lit-plugin",
```
43081j, line 35: finally got rid of that pesky thing :tada: 

just to check i understand properly - we now install everything as normal, then bundle the plugin sources _with_ dependencies and thats why we can get rid of the copylink/install:safe stuff. but we still copy tsc etc since vsc wants it.

rictic: Yes, with the caveat that it's not quite install as normal because we let lerna do its thing, so there's a few symlinks flying around between the packages

And to build the vscode plugin we create a directory at `./packages/vscode-lit-plugin/built` that contains only and exactly those files that are needed, which is very few because we're also using esbuild to bundle almost all dependencies. Our tests of the plugin, as well as the package script, only uses that `built` directory

packages/vscode-lit-plugin/package.json
```diff
@@ -116,26 +117,59 @@
...
+			]
+		},
+		"make-built-dir": {
+			"dependencies": [
+				"../ts-lit-plugin:build",
+				"bundle"
+			],
+			"files": [
+				"../ts-lit-plugin/package.json",
+				"./package.json",
```
aomarks, line 141: Very nit, but there's a mix of globs with and without the `./` prefix, which is a no-op. Personally I think I find it easier to read without the prefix (so just `package.json` instead of `./package.json`), though main comment is just that it'd be nice to be consistent.

aomarks (approved):
Looks reasonable! This is the first time I've really looked at this package so I'm probably unlikely to have caught anything subtle.

packages/vscode-lit-plugin/copy-to-built.js
```diff
@@ -0,0 +1,48 @@
+const { copy, mkdirp, writeFile } = require("fs-extra");
```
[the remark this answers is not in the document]
rictic: Not currently, though we may add them in a future change

PR #246, opened 2022-04-15
Only copy the typescript files that we need

rictic:
This drops our bundle size from 15MB to 7.41MB

### DIFF
--- a/packages/vscode-lit-plugin/copy-to-built.js
+++ b/packages/vscode-lit-plugin/copy-to-built.js
@@ -13,7 +13,10 @@ const { copy, mkdirp, writeFile } = require("fs-extra");
 async function main() {
 	// We don't bundle the typescript compiler into ./built/bundle.js, so we need
 	// a copy of it.
-	await copy("./node_modules/typescript", "./built/node_modules/typescript");
+	await mkdirp("./node_modules/typescript/lib");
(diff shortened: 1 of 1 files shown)


packages/vscode-lit-plugin/copy-to-built.js
```diff
@@ -13,7 +13,10 @@ const { copy, mkdirp, writeFile } = require("fs-extra");
 async function main() {
 	// We don't bundle the typescript compiler into ./built/bundle.js, so we need
 	// a copy of it.
-	await copy("./node_modules/typescript", "./built/node_modules/typescript");
+	await mkdirp("./node_modules/typescript/lib");
+	await copy("./node_modules/typescript/package.json", "./built/node_modules/typescript/package.json");
+	await copy("./node_modules/typescript/lib/typescript.js", "./built/node_modules/typescript/lib/typescript.js");
```
justinfagnani, line 18: we don't need any of the libs?

rictic: It seems like no, we don't

I wish that we didn't need typescript here at all, because we only use it in a marginal way via lit analyzer to find html templates to find styles to find colors in those styles to tell vscode about them

justinfagnani (approved):

PR #249, opened 2022-04-19
Move all bundling into vscode-lit-plugin

rictic:
This way downstream users of ts-lit-plugin are free to bundle it with their preferred settings.


packages/vscode-lit-plugin/package.json
```diff
@@ -122,13 +122,16 @@
 			"dependencies": [
 				"../lit-analyzer:build"
 			],
-			"command": "node esbuild.script.js",
+			"command": "node esbuild.script.mjs",
 			"files": [
 				"src/**/*",
-				"esbuild.script.js"
+				"../ts-lit-plugin/src/**/*",
+				"../ts-lit-plugin/index.js",
```
rictic, line 129: @aomarks is this kosher, or should I put this script at the toplevel of the monorepo?

aomarks: Should be fine, input files can reach out of the package. It's only output files that currently can't.

packages/vscode-lit-plugin/esbuild.script.mjs
```diff
@@ -0,0 +1,33 @@
...
+  target: "es2017",
+  format: "cjs",
+  color: true,
+  mainFields: ["module", "main"]
+});
+
+try {
+  await Promise.all([buildUs, buildTsLitPlugin]);
+} catch (e) {
+  process.exit(1);
```
aomarks, line 32: Log the error?

aomarks (approved):

packages/vscode-lit-plugin/esbuild.script.mjs
rictic: This is the recommended style from the esbuild docs, the error has some useful info but it's mostly a retread of the error that's already been logged at that point.

On the other hand... simplicity is nice, and this is hella fast already. Switched to just awaiting directly.

aomarks (approved):

justinfagnani (approved):

PR #277, opened 2022-09-21
Bump typescript version.

rictic:


justinfagnani:
What about updating the versions for the tests? Should we add 4.3 or 4.7?

rictic:
Yeah, we need to update the unit tests too, but this is just for the vscode integration tests

justinfagnani (approved):

PR #288, opened 2022-11-28
Fix matching of Clutz-generated types against allowed type names.

bicknellr:
Currently, the original name of the Clutz-generated type is being matched against the names of the allowed types. This PR updates this to match against the normalized version.


rictic (approved):
Good catch!

PR #303, opened 2023-04-12
Add support for TypeScript 5.0

bicknellr:
- Installs TypeScript 5.0 as a dev dependency and adds it to the list of versions that are tested.
- Updates web-component-analyzer to `1.1.7`, which added support for TypeScript 5.0.
- Fixes a few dependencies that no longer work and works around a TS compiler API deprecation.


bicknellr:
cc @rictic, here's what I have so far. The tests pass but it breaks while trying to package the plugin, which I haven't looked into at all.

rictic:
Superceded by #305

PR #305, opened 2023-06-13
Update web-component-analyzer dependency to fix ts5.x support

43081j:
* Update WCA
* Update typescript to match WCA's version

this should fix the various problems people are having

cc @rictic 

FYI i couldn't write an extra test that fails before this change as i'd need to change the repo's typescript to 5.x, and that breaks the build thanks to a mismatch of types

we could just publish a prerelease and test it manually

let me know what you think

### DIFF
--- a/.github/workflows/test_vscode.yml
+++ b/.github/workflows/test_vscode.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Install Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 16.x
+          node-version: 20.x
       - run: npm ci
       # Test normally on Mac
       - run: npm run test:packaged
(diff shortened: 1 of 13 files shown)


rictic:
Weird that the vscode downloader fails on windows with Node >16, but this modernizes the rest of the tests' node versions

rictic (approved):

artemyarulin:
Thanks for your PR - can confirm that using `"lit-analyzer": "^2.0.0-pre.3",` fixes problem with running `lit-analyzer` as a command line utility. Unfortunately TypeScript plugin and VS code extension doesn't work still, but I assume it will be fixed later.

Thanks you!

43081j:
yup we still need to publish the new version of the vscode extension, so that will still have the problems i think for now

but peter is looking into it so should be up soon

great to know the CLI part works though 🙏

rictic:
Sorry it took a bit of extra time, but the VSCode extension has now been
released!

PR #309, opened 2023-06-21
Support TypeScript 5.1

rictic:
@43081j

### DIFF
--- a/packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/html-document.ts
+++ b/packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/html-document.ts
@@ -12,7 +12,8 @@ export class HtmlDocument extends TextDocument {
 
 	htmlAttrAreaAtOffset(offset: DocumentOffset | DocumentRange): HtmlNode | undefined {
 		return this.mapFindOne(node => {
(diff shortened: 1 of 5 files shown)


packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/html-document.ts
```diff
@@ -12,7 +12,8 @@ export class HtmlDocument extends TextDocument {
 
 	htmlAttrAreaAtOffset(offset: DocumentOffset | DocumentRange): HtmlNode | undefined {
 		return this.mapFindOne(node => {
-			if (offset > node.location.name.end && intersects(offset, node.location.startTag)) {
+			const offsetNum = typeof offset === "number" ? offset : offset.end;
```
43081j, line 15: does 5.1 change what a `DocumentOffset` is? or was this a bug before? :D

43081j (approved):

packages/lit-analyzer/src/lib/analyze/parse/document/text-document/html-document/html-document.ts
43081j: nice find in that case!

PR #320, opened 2023-09-05
Fixes #319

### LINKED ISSUE #319: [vscode-lit-plugin] Non-hex color formats aren't recognized for color picker
Related to #30 

Currently, the only colors that can be matched and that can use the color picker are hex based: https://github.com/runem/lit-analyzer/blob/9659154904fd2010de3749fc5e904e8c3dfe684b/packages/vscode-lit-plugin/src/color-provider.ts#L117

With an HTML file:
image

With a TS file:
image

Reproduce:

```ts
import { LitElement, css, html } from 'lit'
import { customElement } from 'lit/decorators.js'

@customElement('foo-element')
export class FooElement extends LitElement {
  render() {
    return html`<p>hello</p>`;
  }

  static styles = css`
    .foo {
      background: white;
      background-color: rgb(234, 238, 242);
      color: hsl(120deg, 75%, 25%);
      color: hsla(120deg, 75%, 25%, 0.5);
      color: #fff;
    }
  `
}

```

au5ton:
This adds support for CSS colors in the following formats:
- rgb(...)
- rgba(...)
- hsl(...)
- hsla(...)

Fixes #319

### DIFF
--- a/packages/vscode-lit-plugin/src/color-provider.ts
+++ b/packages/vscode-lit-plugin/src/color-provider.ts
@@ -1,85 +1,15 @@
 import * as vscode from "vscode";
+import { HSLA, RGBA } from "./color.js";
 
 /**
  * Regex to match colors in a string
  */
-const COLOR_HEX_REGEX = /#[0-9a-fA-F]+/gi;
+//const COLOR_HEX_REGEX = /#[0-9a-fA-F]+/gi;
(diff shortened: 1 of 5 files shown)


au5ton:
@runem What do you think? Is there anything I can do to improve this before it can be accepted?

PR #333, opened 2023-12-07
Fixes #313 by correctly using projects moduleResolution strategy if configured

### LINKED ISSUE #313: [vscode-lit-plugin] Typescript 5.1 Missing import warning for <my-element> doesn't work
@rictic Missing import for <element> still doesn't work with latest typescript version

benjamind:
* Updates the repo to typescript 4.8.4 - this is the minimum tested TS version anyway, and supports the correct use of `getNodeForUsageLocation` to properly determine `ModuleKind` for resolving modules.
* Updates use of `resolveModuleNameFromCache` to properly pass the right mode argument to resolve in more situations correctly.
* Updates required and default `compilerOptions` to not overwrite the project under tests tsconfig options if provided. Will still default to the original behavior, but will now obey whats in the `tsconfig.json` if its specified.
* Updates lots of package-locks which didn't seem to be up to date in `main` and were updated automatically.

### DIFF
--- a/packages/lit-analyzer/src/lib/analyze/parse/parse-dependencies/visit-dependencies.ts
+++ b/packages/lit-analyzer/src/lib/analyze/parse/parse-dependencies/visit-dependencies.ts
@@ -158,8 +158,8 @@ function emitDirectModuleImportWithName(moduleSpecifier: string, node: Node, con
 		result = (context.program as any)["getResolvedModuleWithFailedLookupLocationsFromCache"](moduleSpecifier, fromSourceFile.fileName);
 	} else {
 		const cache = (context.program as MaybeModernProgram).getModuleResolutionCache?.();
-		let mode: tsModule.ModuleKind.CommonJS | tsModule.ModuleKind.ESNext | undefined = undefined; // tsModule.ModuleKind.ESNext;
-		if ((context.ts.isImportDeclaration(node) && !node.importClause?.isTypeOnly) || (context.ts.isExportDeclaration(node) && !node.isTypeOnly)) {
+		let mode: tsModule.ModuleKind.CommonJS | tsModule.ModuleKind.ESNext | undefined = undefined;
+		if (context.ts.isImportDeclaration(node) || context.ts.isExportDeclaration(node)) {
 			if (node.moduleSpecifier != null && context.ts.isStringLiteral(node.moduleSpecifier) && context.ts.isSourceFile(node.parent)) {
 				mode = tsModule.getModeForUsageLocation(fromSourceFile, node.moduleSpecifier);
 			}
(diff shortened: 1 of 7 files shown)


packages/lit-analyzer/src/lib/analyze/parse/parse-dependencies/visit-dependencies.ts
```diff
@@ -158,8 +158,15 @@ function emitDirectModuleImportWithName(moduleSpecifier: string, node: Node, con
 		result = (context.program as any)["getResolvedModuleWithFailedLookupLocationsFromCache"](moduleSpecifier, fromSourceFile.fileName);
 	} else {
 		const cache = (context.program as MaybeModernProgram).getModuleResolutionCache?.();
+		let mode: tsModule.ModuleKind.CommonJS | tsModule.ModuleKind.ESNext | undefined = undefined; // tsModule.ModuleKind.ESNext;
+		if ((context.ts.isImportDeclaration(node) && !node.importClause?.isTypeOnly) || (context.ts.isExportDeclaration(node) && !node.isTypeOnly)) {
```
rictic, line 162:
```ts
!node.importClause?.isTypeOnly
```

Should this instead be:

```ts
(node.importClause && !node.importClause.isTypeOnly)
```

Because this import doesn't have an import clause, but it shouldn't be treated like a type only import:

```ts
import 'foo';
```

Hm, actually, looking at the docs for getModeForUsageLocation, I think we should just pass the node itself and not have any of these checks, including the typeOnly check:

```ts
/**
 * Calculates the final resolution mode for a given module reference node. This is generally the explicitly provided resolution mode, if
 * one exists, or the mode of the containing source file. (Excepting import=require, which is always commonjs, and dynamic import, which is always esm).
 * Notably, this function always returns `undefined` if the containing file has an `undefined` `impliedNodeFormat` - this field is only set when
 * `moduleResolution` is `node16`+.
 * @param file The file the import or import-like reference is contained within
 * @param usage The module reference string
 * @returns The final resolution mode of the import
 */
```

benjamind: Yeah I tried to do that without the checks but it broke in TS 4.8.3 while passed in later TS versions. Not sure why exactly.

But yeah this conditional I just copied from elsewhere in this file so I'm not sure its the right one exactly.

rictic: How does it break in 4.8.3?

rictic (approved):

rictic:
Thanks for this great change! I'll stamp a release for this on Monday

benjamind:
Nice! Thanks @rictic hopefully this solves a bunch of problems for folks.

rictic:
Published as v2.0.2 of lit-analyzer on npm and v1.4.1 of runem/lit-plugin in the vscode extension marketplace

PR #335, opened 2023-12-12
Adds integration test for no-missing-import

benjamind:
This adds an integration test for the vscode plugin to demonstrate the failure case for no-missing-import.

It loads a file with a missing import, and then adds the line into the editor for the import. This should fix all errors in the file, but instead we see a diagnostic for the other element in the page. It seems the extension is only resolving one import at a time.

This could be (and most likely is) a bug in lit-analyzer package itself, but only manifests in the extension where our TS state is dynamic. I'm guessing its a caching issue?


AndrewJakubowicz:
Thank you so much for your test. It was included in https://github.com/runem/lit-analyzer/pull/338 with the fix.

PR #337, opened 2024-01-03
[infra] update vscode-test to fix access issues when locally testing

AndrewJakubowicz:
### Context

Discovered this when trying to locally test and develop the analyzer. This fix is from my issue filed on vscode-test.

Relevant issue: https://github.com/microsoft/vscode-test/issues/254

This may also fix GitHub actions that had errors.

### Risk

Risk should be minimal as this updates a test only dependency.


AndrewJakubowicz:
I do not have permissions to request review, so cc @rictic 
Thank you!

rictic (approved):

PR #338, opened 2024-01-05
fix no-missing-import getting confused on incremental file changes

AndrewJakubowicz:
### Context

See the following video to preview the issue:

https://github.com/runem/lit-analyzer/assets/15080861/c9eb6c43-1f1c-4b14-a012-79622a29a7ac

This screen recording captures the integration test created by @benjamind in https://github.com/runem/lit-analyzer/pull/335

### Fix

This was a rough issue to debug. Ultimately the TypeScript cache seems to sometimes not contain the files required to resolve the module specifier string. The fix is to handle this case by resolving the module specifier without using the cache.

This graceful fallback causes the integration test to pass. I also manually verified that the fix works.

### Other changes

I also did a small amount of cleanup removing some commented out code and tiny changes as I had to go pretty deep to find this issue.

### DIFF
--- a/packages/lit-analyzer/src/lib/analyze/default-lit-analyzer-context.ts
+++ b/packages/lit-analyzer/src/lib/analyze/default-lit-analyzer-context.ts
@@ -285,6 +285,6 @@ export class DefaultLitAnalyzerContext implements LitAnalyzerContext {
 
 		// Build a graph of component dependencies
 		const res = parseDependencies(file, this);
-		this.dependencyStore.importedComponentDefinitionsInFile.set(file.fileName, res);
+		this.dependencyStore.absorbComponentDefinitionsForFile(file, res);
 	}
 }
--- a/packages/lit-analyzer/src/lib/analyze/parse/parse-dependencies/visit-dependencies.ts
+++ b/packages/lit-analyzer/src/lib/analyze/parse/parse-dependencies/visit-dependencies.ts
@@ -168,6 +168,16 @@ function emitDirectModuleImportWithName(moduleSpecifier: string, node: Node, con
 		if (cache != null) {
 			result = context.ts.resolveModuleNameFromCache(moduleSpecifier, node.getSourceFile().fileName, cache, mode);
 		}
+		if (result == null) {
+			// Result could not be found from the cache, try and resolve module without using the
+			// cache.
+			result = context.ts.resolveModuleName(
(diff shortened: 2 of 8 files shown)


packages/lit-analyzer/src/lib/analyze/parse/parse-dependencies/visit-dependencies.ts
```diff
@@ -168,6 +168,16 @@ function emitDirectModuleImportWithName(moduleSpecifier: string, node: Node, con
...
+		if (result == null) {
+			// Result could not be found from the cache, try and resolve module without using the
+			// cache.
+			result = context.ts.resolveModuleName(
+				moduleSpecifier,
+				node.getSourceFile().fileName,
+				context.program.getCompilerOptions(),
+				context.ts.createCompilerHost(context.program.getCompilerOptions())
+			);
+		}
```
AndrewJakubowicz, line 180: This fixes the issue. The rest of the PR is refactor & tests

benjamind: Genius. Nice find.

packages/lit-analyzer/src/lib/analyze/default-lit-analyzer-context.ts
```diff
@@ -285,6 +285,6 @@ export class DefaultLitAnalyzerContext implements LitAnalyzerContext {
 
 		// Build a graph of component dependencies
 		const res = parseDependencies(file, this);
-		this.dependencyStore.importedComponentDefinitionsInFile.set(file.fileName, res);
+		this.dependencyStore.absorbComponentDefinitionsForFile(file, res);
```
rictic, line 288: What's the difference between these methods?

AndrewJakubowicz: There is no difference except it looked like the intent was that `importedComponentDefinitionsInFile` should be `private`, with `absorbComponentDefinitionsForFile` being the public method to do this logic.

You can see the `dependencyStore` class defined here: https://github.com/runem/lit-analyzer/pull/338/files#diff-faaba94b3ffda920023b5e0ff3b54f7137199117142a418b7bd79df7e8951c5eR6-R9

rictic (approved):
Awesome work!

Issue #317, opened 2023-08-02
Update `no-nullable-attribute-binding` rule for null attribute


xiaohk:
It seems `ifDefined` now supports both `undefined` and `null` (https://github.com/lit/lit/issues/1927)

This rule should be updated.

https://github.com/runem/lit-analyzer/blob/9659154904fd2010de3749fc5e904e8c3dfe684b/packages/lit-analyzer/src/lib/rules/no-nullable-attribute-binding.ts#L26-L41

PR #339, opened 2024-01-05
Update no-nullable-attribute-binding fix for null attribute

AndrewJakubowicz:
### Context

Fixes: https://github.com/runem/lit-analyzer/issues/317

In versions of Lit >2, `ifDefined` can handle `null` and `undefined` so the explicit `null` check is no longer required.

### Change

Instead of recommending `ifDefined(expr === null ? undefined : expr)`, instead recommend `ifDefined(expr)` for both `undefined` and `null`.

### Test

Added message tests to ensure that type detection works correctly, and that the diagnostic message correctly identify either `null` or `undefined`. This will make it easier to catch regressions when moving to `isTypeAssignableTo` in the future.


AndrewJakubowicz:
Ready for review, cc/ @rictic

rictic (approved):

rictic:
Great change!

Issue #283, opened 2022-10-28
Wrongly reports "This tag isn't closed" on <p> <div></div></p>


nmattia:
The CLI version of `lit-analyzer` will incorrectly report a `<p>` tag starting with a space as unclosed:

```
./src/file.ts

    This tag isn't closed.
    92:  <p> <div></div></p>
    no-unclosed-tag
```

Removing the whitespace fixes the issue.

AndrewJakubowicz:
I think the issue here is that the workaround of `removing the whitespace` should not resolve the issue. The error message can also be made clearer.

Regarding the example html provided, both with and without the whitespace results in DOM being created that doesn't reflect what was authored.

Due to the paragraph tag allowing ["tag omission"](https://html.spec.whatwg.org/multipage/dom.html#concept-element-tag-omission), the `<div>` will result in the `<p>` being self-closed.
When parsed by the browser (and `parse5`), the html: `<p> <div></div></p>` or `<p><div></div></p>` creates the following DOM:

```html
<p> </p><div></div><p></p>
```

or 

```html
<p></p><div></div><p></p>
```


In both cases, with or without the space, the html as written won't reflect what is rendered in the DOM.


The specification reference for why this happens is described here: https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element

Relevant excerpt (with large list of tags omitted):

> "A [p](https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element) element's [end tag](https://html.spec.whatwg.org/multipage/syntax.html#syntax-end-tag) can be omitted if the [p](https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element) element is immediately followed by a ... [div](https://html.spec.whatwg.org/multipage/grouping-content.html#the-div-element) ... element"


I think the action item here is that both cases should raise the `This tag isn't closed` error. However maybe a message can be added to warn that this diagnostic is raised in the case where closing tags were automatically added by parse5/browser that were not authored?

PR #340, opened 2024-01-06
Make `isSelfClosed` check more explicit

AndrewJakubowicz:
### Context

This was inspired (and maybe closes) https://github.com/runem/lit-analyzer/issues/283.

This may be the wrong direction for fixing the linked issue, however I wanted to propose a solution after my investigation.

### Current check

The current `no-unclosed-tag` check grabs the source code locations for the tag from `parse5`, and then tries to guess if the tag was closed.

The tag _is marked as self closing_ if the node has **no children** and the `startTag.endOffset === node.endOffset`.

An important note here, is that the source code locations are taken from `parse5`, which behaves similarly to the browser and thus has special parsing rules for various tags.

The edge case can be seen with `<p>` which can be closed implicitly.

```html
<p><div></div><!--becomes:--><p></p><div></div>
<p>Some content<div></div><!--becomes:--><p>Some content</p><div></div>
```

Currently, the first case without content will not raise the `no-unclosed-tag` check, but the second case does raise the `no-unclosed-tag` check.

Why? Because `parse5` correctly creates the DOM `<p>Some content</p><div></div>`, but then there is no reference to an authored closing tag, and the presence of child content fails to mark the tag as self closing.

### Possible change

This PR introduces a possible change where the intent of self closing can be marked by adding `/>`. Therefore implicit closing tags result in the `no-unclosed-tag` diagnostic being raised. To mark a tag as explicit, add the `/>` syntax (which FYI isn't a valid HTML syntax).

### Risk

This will catch more cases, so I'm not very stoked about this idea. I'm also not excited by the general concept of "self-closing" syntax which this PR pushes forward.

### DIFF
--- a/packages/lit-analyzer/src/lib/rules/no-unclosed-tag.ts
+++ b/packages/lit-analyzer/src/lib/rules/no-unclosed-tag.ts
@@ -11,7 +11,11 @@ const rule: RuleModule = {
 		priority: "low"
 	},
 	visitHtmlNode(htmlNode, context) {
-		if (!htmlNode.selfClosed && htmlNode.location.endTag == null) {
+		const {
+			selfClosed,
+			location: { endTag }
+		} = htmlNode;
+		if (!selfClosed && endTag == null) {
 			// Report specifically that a custom element cannot be self closing
 			//   if the user is trying to close a custom element.
 			const isCustomElement = isCustomElementTagName(htmlNode.tagName);
--- a/packages/lit-analyzer/src/test/rules/no-unclosed-tag.ts
+++ b/packages/lit-analyzer/src/test/rules/no-unclosed-tag.ts
@@ -11,3 +11,33 @@ tsTest("Don't report self closed tags", t => {
 	const { diagnostics } = getDiagnostics("html`<img />`", { rules: { "no-unclosed-tag": true } });
 	hasNoDiagnostics(t, diagnostics);
 });
+
+tsTest("Don't report void tags", t => {
+	const { diagnostics } = getDiagnostics("html`<img>`", { rules: { "no-unclosed-tag": true } });
+	hasNoDiagnostics(t, diagnostics);
+});
+
+// The `<p>` tag will be closed automatically if immediately followed by a lot of other elements,
+// including `<div>`.
+// Ref: https://html.spec.whatwg.org/multipage/grouping-content.html#the-p-element
+tsTest("Report unclosed 'p' tag that is implicitly closed via tag omission", t => {
+	const { diagnostics } = getDiagnostics("html`<p><div></div></p>`", { rules: { "no-unclosed-tag": true } });
+	hasDiagnostic(t, diagnostics, "no-unclosed-tag");
+});
+
+tsTest("Report unclosed 'p' tag that is implicitly closed via tag omission containing text content", t => {
+	const { diagnostics } = getDiagnostics("html`<p>Unclosed Content<div></div></p>`", { rules: { "no-unclosed-tag": true } });
+	hasDiagnostic(t, diagnostics, "no-unclosed-tag");
+});
+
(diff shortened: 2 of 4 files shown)


43081j:
maybe we can actually just check that there is an end tag?

for example:

```html
  <p>i never close
```

will have a `startTag` location, and no `endTag` location. 

similarly:

```html
  <p />
```

will behave the same since you can't self-close a `p` tag (parse5 will already be taking this into account i think). so it too won't have an `endTag` and will actually have a child text node (the whitespace following it).

so maybe we can just check `!getNodeSourceLocation(p5Node).endTag`?

i'm not sure why we didn't do that all along... unless i'm missing something 🤔

AndrewJakubowicz:
cc @43081j  and @rictic 

This is ready for review. Thank you James for your exceptional feedback!

43081j:
the CI failure is because one of the tests actually self-closes a custom element FYI 😬 

so it already caught one!

43081j (approved):
looks good to me 👍 

we should probably double check if it does solve the issue you mentioned too. that was more about a space tripping it up i think

AndrewJakubowicz:
Excellent! :D

1. Added a specific test case for `<p> <div></div></p>` (which continues to raise the diagnostic).
2. Fixed the self-closing custom element in the other test.

43081j (approved):

packages/lit-analyzer/src/lib/rules/no-unclosed-tag.ts
```diff
@@ -11,18 +34,18 @@ const rule: RuleModule = {
...
-		if (!htmlNode.selfClosed && htmlNode.location.endTag == null) {
-			// Report specifically that a custom element cannot be self closing
-			//   if the user is trying to close a custom element.
-			const isCustomElement = isCustomElementTagName(htmlNode.tagName);
-
-			context.report({
-				location: rangeFromHtmlNode(htmlNode),
-				message: `This tag isn't closed.${isCustomElement ? " Custom elements cannot be self closing." : ""}`
-			});
+		if (VOID_ELEMENTS.has(htmlNode.tagName.toLowerCase()) || htmlNode.location.endTag != null) {
```
rictic, line 37: Do we warn on an invalid closing element for a void element? Seems like it'd be good to do so. For example, this should be a warning on the `</input>`

```html
<input><div></div></input>
```

packages/lit-analyzer/src/test/rules/no-unclosed-tag.ts
```diff
@@ -7,7 +7,49 @@ tsTest("Report unclosed tags", t => {
 	hasDiagnostic(t, diagnostics, "no-unclosed-tag");
 });
```
rictic, line 9: In an ideal world, we might want to allow:

```html
<p>One
<p>Two
```

Or just:

```html
<p>Hi
```

And we'd want to put warnings on dangling closing tags that don't match up to any opening tags.

All of those may be somewhat tricky to do with how parse5 represents the tree though, requiring us to notice spans in the original text that aren't represented in the parsed tree.

packages/lit-analyzer/src/lib/rules/no-unclosed-tag.ts
AndrewJakubowicz: I think this is a good idea but maybe something we can add incrementally to the `no-unclosed-tag` diagnostic.

I'll make an issue.

packages/lit-analyzer/src/test/rules/no-unclosed-tag.ts
AndrewJakubowicz: These are very tricky, and another good case to add to an issue for followup.

Philosophically though, I wonder if the lit-analyzer should avoid HTML diagnostics and leave that to an HTML specific tool. It may even be worth scoping the `no-unclosed-tag` to only check custom element tags.

Issue #326, opened 2023-10-13
`color-scheme` marked as an invalid css


perkrlsn:
Hello gang! Seems like the [color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) property is marked as invalid.

![image](https://github.com/runem/lit-analyzer/assets/12989781/076aae86-4fdb-4515-a651-72aa61b7c323)

Issue #284, opened 2022-10-28
no-invalid-css fails on inlined animation keyframes


nmattia:
The `no-invalid-css` rule gets triggered by (as far as I can tell) perfectly fine CSS animation keyframes:

```
  <style>
    @keyframes test {
        0% {
            background-color: transparent;
        }
    }
  </style>
```

gives:

```

./src/file.ts

    { expected
    286:  @keyframes test {
    no-invalid-css
```

Issue #259, opened 2022-05-16
lit-plugin complains about media queries in <style> tags


WickyNilliams:
consider a template like:

```js
html`
  <style>
    @media (min-width: 768px) { /* (1) */
      body {
        color: red;
      } /* (2) */
    } 
  </style>
`
```

lit-plugin reports the following errors:

1. `{ expected lit-plugin(no-invalid-css)`
2. `at-rule or selector expected lit-plugin(no-invalid-css)`

despite this being valid css

ba55ie:
Why aren't you using the `css` tagged template literal?

```
static styles = css`
   @media (min-width: 768px) {
      body {
        color: red;
      }
    } 
`;
```

https://lit.dev/docs/components/styles/

WickyNilliams:
The example given is a reduced test case for illustrative purposes only. I have experienced this issue in tests where I'm writing fixtures, and also cases where you need dynamic styles (as described here https://lit.dev/docs/components/styles/#style-element). So `css` doesn't cut it

Issue #146, opened 2020-11-05
no-invalid-css flagged for the css property content-visibility


ARuzek:
content-visibility is a fairly new css property. Please add validation. 

ref: https://web.dev/content-visibility/

KeithHenry:
`contain-intrinsic-size` has the same issue

PR #352, opened 2024-02-02
Update CSS and HTML Language Service versions

### LINKED ISSUE #314: lit-plugin doesn't recognize certain CSS properties.
lit-plugin may not have updated its CSS property library, this is an omission I found currently.

`Unknown property: 'accent-color' lit-plugin(no-invalid-css)`

https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color

JHonaker:
This is a pretty simple PR. I updated the CSS and HTML Language Service packages to their latest release. The current locked versions are from mid 2020, and there have been many new features like the `accent-color` CSS attribute that prompted this work.

This fixes #314, most likely #146, potentially #326, #132, #284, and #259 (but I haven't tested it).

I also updated the TypeScript version to 5.3.3, and added the appropriate compatibility checks to the test suite for 5.2 and 5.3.

### DIFF
--- a/packages/lit-analyzer/package-lock.json
+++ b/packages/lit-analyzer/package-lock.json
@@ -15,7 +15,7 @@
 				"fast-glob": "^3.2.11",
 				"parse5": "5.1.0",
 				"ts-simple-type": "~2.0.0-next.0",
-				"vscode-css-languageservice": "4.3.0",
+				"vscode-css-languageservice": "6.2.12",
 				"vscode-html-languageservice": "3.1.0",
 				"web-component-analyzer": "^2.0.0"
 			},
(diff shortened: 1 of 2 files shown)


AndrewJakubowicz (approved):
Awesome! Thank you!

rictic:
It looks like the TypeScript type incompat issue is breaking on CI. Updating TypeScript is good to do to ensure that we're compatible across all different versions, but unless we find and fix an incompatibility it actually doesn't impact the user because we use the locally installed version of TypeScript in the user's code.

IMO let's do the TS update as a separate change, because to keep the build green it needs to be coordinated across a couple repos

JHonaker:
Yea, I didn’t see anything in the tests locally, but it’s just the last commit. Very easy to rollback. I’ll do it when I get on my computer later.

JHonaker:
Okay, @rictic I rolled back the TypeScript change. We should be good to go now!

rictic (approved):

klasjersevi:
Looking forward to this, any idea when it will be released?

fdendorfer:
@runem Sorry for the ping, but is a release scheduled for this. Latest Stylelint standard ruleset expects some of these new css features which causes them to show up as errors in the IDE

klasjersevi:
@rictic @runem Any updates on a release for this?

SuperLlama88888:
@rictic Really sorry for the ping. Any ETA on this being released?

SuperLlama88888:
An up-to-date fork is here: https://marketplace.visualstudio.com/items?itemName=jackolope.lit-analyzer-plugin
Source code: https://github.com/JackRobards/lit-analyzer/tree/main/packages/vscode-lit-plugin

Issue #253, opened 2022-04-28
no-nullable-attribute-binding not working properly and cannot turn off


ggecy:
Hello, in the following variable defined (and computed later on):
```
let className: string | null = null
```
and then used in a template
```
result = render(html`<span class="${ifDefined(className)}">${model.item.values[columnIndex]}</span>`, root);
```
which shows the following error for class attribute: `This attribute binds the type 'string | null' which can end up binding the string 'null'. Use the 'ifDefined' directive and strict null check?lit-plugin(no-nullable-attribute-binding)(2314)`.

This shouldn't be an error according to official lit element documentation and implementation of ifDefined directive - it works with both `undefined` and `null` values and ignores the attribute : `For AttributeParts, sets the attribute if the value is defined and removes the attribute if the value is undefined (undefined or null)`

I have also tried setting the `no-nullable-attribute-binding` to off in settings yet the error is still shown.

jesseditson:
Just giving this a quick bump as I've otherwise been finding this lib incredibly helpful but this bug has really cramped my CI.

nathanbrnrd:
This is definitely polluting my IDE with error messages that are not relevant.
A quick patch that I use is to install the ts-lit-plugin and change error level in the tsconfig.json:
```
"plugins": [
        {
          "name": "ts-lit-plugin",
          "rules": {
            "no-nullable-attribute-binding": "warning",
          }
        }
      ]
```

Note that setting the rule value to "off" doesn't work

PR #360, opened 2024-06-11
Add a test that the no-nullable-attribute-binding fix suggestion actually works.

jyasskin:
This is a test to go with whatever fix is developed for #253. It can't be submitted as-is since it currently fails. It also demonstrates a workaround for the bug: use `??nothing` instead of `ifDefined`.


justinfagnani:
Wow, thanks for writing a test!

PR #364, opened 2024-07-27
Exclude symbols when checking binding types

### LINKED ISSUE #207: Problem with custom directives (no-incompatible-type-binding)
I have been trying to write a simple custom directive which simplifies some use-cases for `ifDefined`:

```js
// instead of
html`<input pattern=${ifDefined(isNumber ? "[0-9]*" : undefined)} />`

// i would like to do this
html`<input pattern=${cond(isNumber, "[0-9]*")} />`
```

Cribbing from ifDefined, I came up with this directive:

```ts
const cond = <T>(condition: boolean, value: T) => (condition ? value : nothing)
```

This directive works just fine at runtime, but the vscode plugin complains:

```
Type 'Symbol(undefined) | "[0-9]*"' is not assignable to 'string'
```

So I also tried creating a directive as recommended in the lit docs:

```ts
class ConditionalDirective extends Directive {
  constructor(partInfo: PartInfo) {
    super(partInfo)
    if (partInfo.type !== PartType.ATTRIBUTE) {
      throw new Error("The `cond` directive must be on attributes")
    }
  }

  render(condition: boolean, value: string) {
    return condition ? value : nothing
  }
}

// Create the directive function
cond = directive(ConditionalDirective)
```

But now i get a different error:

```
You are binding a non-primitive type 'DirectiveResult<{ prototype: any; new(partInfo: ChildPartInfo | AttributePartInfo | ElementPartInfo) => ConditionalDirective; }>'. 
This could result in binding the string "[object Object]". Use '.' binding instead?
```

Again, the directive itself works fine at runtime... 

I tried following the advice in the error message but then I get a different error again:

```
Type 'DirectiveResult<{ prototype: any; new(partInfo: ChildPartInfo | AttributePartInfo | ElementPartInfo) => ConditionalDirective; }>' is not assignable to 'string'
```

Am I doing something wrong here, or is this a bug with lit-analyzer?

### LINKED ISSUE #251: Treat noChange and nothing as any type in binding
The noChange and nothing values have special meaning in a lit binding, and they are legal to use everywhere, as they directly specify behavior rather than an ordinary value to be set/rendered

### LINKED ISSUE #316: Conditionally rendering nothing on an aria attribute produces an error when it shouldn't
The code below is an example taken from an element class that demonstrates a scenario that triggers the following error on the `aria-current` attribute for the link element.

```
Type '"page" | Symbol(undefined)' is not assignable to '"page" | "step" | "location" | "date" | "time" | "true" | "false"'lit-plugin(no-incompatible-type-binding)(2304)

```

```typescript
#renderLinks(): TemplateResult {
  return html`
    <ul>
      ${this.#menuItems.map((link) => {
      return html`
        <li>
          <a href="${link.href.pathname}" aria-current="${this.#isCurrentPage(link.href) || nothing }">${link.text}</a>
        </li>
      `})
    }
    </ul>
  `;
}

#isCurrentPage(url: URL){
  if(url.pathname === window.location.pathname){
    return 'page';
  } else {
    return undefined;
  }
}
```

This should not trigger an error as per these guidelines: https://lit.dev/docs/templates/conditionals/#conditionally-rendering-nothing

SegaraRai:
This PR excludes ES symbol types from unions when checking bindings.
This ensures that `nothing` and `noChange` are ignored in type checking.

We remove all symbol types because it is not possible to get the details of symbol types.
This will produce false negatives when other symbols are mixed in, but I consider such situations to be rare.

Closes #207. Closes #251. Closes #316.

This PR does not resolve the issue of `ifDefined` not supporting `null` (even though Lit v3 does).
It will probably be resolved when we stop treating `ifDefined` specially, or when #296 is merged.
The user can use either a custom directive or `value ?? nothing` instead of `ifDefined` to workaround the issue.

### DIFF
--- a/packages/lit-analyzer/src/lib/rules/util/type/extract-binding-types.ts
+++ b/packages/lit-analyzer/src/lib/rules/util/type/extract-binding-types.ts
@@ -50,6 +50,10 @@ export function extractBindingTypes(assignment: HtmlNodeAttrAssignment, context:
 		typeB = directiveType;
 	}
 
+	// Handle `nothing` and `noChange` symbols
+	// Since it's not possible to check the details of a symbol due to various restrictions (it's treated as a `unique symbol` or `Symbol()`), all symbols are excluded.
+	typeB = excludeSymbolsFromUnion(typeB);
+
 	// Cache the result
 	const result = { typeA, typeB };
 	cache.set(assignment, result);
@@ -80,6 +84,17 @@ export function inferTypeFromAssignment(assignment: HtmlNodeAttrAssignment, chec
 
 			return { kind: "STRING" } as SimpleTypeString;
 	}
+}
+
+function excludeSymbolsFromUnion(type: SimpleType): SimpleType {
+	if (type.kind !== "UNION") {
+		return type;
+	}
+
+	return {
+		...type,
+		types: type.types.filter(t => t.kind !== "ES_SYMBOL" && t.kind !== "ES_SYMBOL_UNIQUE")
+	};
 }
 
 /**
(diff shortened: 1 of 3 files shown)


stefanpearson:
Is there any update on merging this? It would be great to have this merged and published, and I'm surprised it hasn't been resolved already!

Binding `nothing` to an attribute is a common thing that the Lit docs recommends when it needs to be omitted, yet my IDE is riddled with red squiggly lines everywhere complaining about it.

PR #400, opened 2026-01-14
fix: support generic args

### LINKED ISSUE #149: Generic Component Declarations
I was creating some custom elements from a generic base class using LitElement and came to some odd problems where the analyzer didn't type-check generic-related properties.

A simplified version of what I was doing:

```ts
export class GenericElement<T> extends LitElement {
	@property() key!: keyof T
}

declare global {
	interface HTMLElementTagNameMap {
		'generic-specific': GenericElement<{ id: number, name: string }>
	}
}
```

First I tried to do it the vanilla way and type checking works fine:
```ts
const element = document.createElement('generic-specific')
element.key = 'what' // ✅ TS2322: Type '"what"' is not assignable to type '"id" | "name"'.
```

However, for some reason, it doesn't work in tagged html templates:
```ts
export const someTemplate = () => html`
        <!-- ❌ Not type-checked -->
	<generic-specific key='what??'></generic-specific>
`
```

I guess the analyzer doesn't analyze the declared generic part in the `HTMLElementTagNameMap`? because it thinks the key is of type `keyof T` instead of `typeof { id: number, name: string }`.

Would be glad to know what I'm missing or to know if there are work-arounds for now.

OrKoN:
This is some support for generics although probably not exhaustive.

Disclaimer: this is AI-assisted (except for test cases) since I do not know TSC API that well.

Closes https://github.com/runem/lit-analyzer/issues/149


OrKoN:
cc @runem in case you or other contributors have a chance to take a look

